PR #4, opened 2015-02-06
Originated channel is now added to a bridge.

TACambron:
The program checks for an available bridge. If a bridge is available,
then that bridge is reused. If no bridges are active, a new bridge
is created. Once the bridge is created, a channel is originated to
a designated endpoint, and upon answering the call, the channel is
added to the bridge.

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,3 +1,4 @@
+/*jslint node: true */  
 'use strict';
 
 var ari = require('ari-client');
@@ -6,10 +7,17 @@ var sla = require('./lib/sla.js');
 
 ari.connect('http://127.0.0.1:8088', 'user', 'pass', clientLoaded);
 
-function clientLoaded (err, client){
-  client.start('hello');
+/**
+ * Starts Stasis app 'sla' and initiates SLA application after ARI connection.
+ * @param {Object} err - Error message.
+ * @param {Object} client - Object that contains information from the ARI 
+ *   connection.
+ */
+function clientLoaded (err, client) {
+  client.start('sla');
   var outgoing = client.Channel();
-  sla(client, outgoing, function(){
-    console.log('done');
+  sla(client, function() {
+    console.log('Application Completed');
+    process.exit(0);
   });
-}
+};
--- a/app.js
+++ b/app.js
@@ -16,15 +16,14 @@ ari.connect('http://127.0.0.1:8088', 'user', 'pass', clientLoaded);
 function clientLoaded (err, client) {
   if (err) {
     throw err;
-  } else {
-    client.start('sla');
-    sla(client, function(err) {
-      if (err !== null) {
-        throw err;
-      } else {
-        console.log('Application Completed');
-        process.exit(0);
-      }
-    });
   }
+  client.start('sla');
+  sla(client, function(err) {
+    if (err !== null) {
+      throw err;
+    } else {
+      console.log('Application Completed');
+      process.exit(0);
+    }
+  });
 }
--- a/app.js
+++ b/app.js
@@ -14,14 +14,17 @@ ari.connect('http://127.0.0.1:8088', 'user', 'pass', clientLoaded);
  *   connection.
  */
 function clientLoaded (err, client) {
-  client.start('sla');
-  sla(client, function(err) {
-    if (err != null) {
-      console.error(err);
-      process.exit(1);
-    } else {
-      console.log('Application Completed');
-      process.exit(0);
-    }
-  });
-};
+  if (err) {
+    throw err;
+  } else {
+    client.start('sla');
+    sla(client, function(err) {
+      if (err !== null) {
+        throw err;
+      } else {
+        console.log('Application Completed');
+        process.exit(0);
+      }
+    });
+  }
+}
--- a/app.js
+++ b/app.js
@@ -9,10 +9,9 @@ var Q = require('q');
 var connect = Q.denodeify(ari.connect);
 
 connect('http://127.0.0.1:8088', 'user', 'pass')
-  .then(function (client) {
-    clientLoaded(client);
-  })
-  .catch(errHandler);
+  .then(clientLoaded)
+  .catch(errHandler)
+  .done();
 
 /**
  * Starts Stasis app 'sla' and initiates SLA application after ARI connection.
@@ -24,9 +23,7 @@ function clientLoaded (client) {
   sla(client)
     .then(console.log)
     .catch(errHandler)
-    .finally(function () {
-      process.exit(0);
-    });
+    .done();
 }
 
 /**
--- a/app.js
+++ b/app.js
@@ -15,9 +15,13 @@ ari.connect('http://127.0.0.1:8088', 'user', 'pass', clientLoaded);
  */
 function clientLoaded (err, client) {
   client.start('sla');
-  var outgoing = client.Channel();
-  sla(client, function() {
-    console.log('Application Completed');
-    process.exit(0);
+  sla(client, function(err) {
+    if (err != null) {
+      console.error(err);
+      process.exit(1);
+    } else {
+      console.log('Application Completed');
+      process.exit(0);
+    }
   });
 };
--- a/app.js
+++ b/app.js
@@ -7,4 +7,9 @@ var sla = require('./lib/sla.js');
 ari.connect('http://127.0.0.1:8088', 'user', 'pass', clientLoaded);
 
 function clientLoaded (err, client){
-    console.log(sla());}
+  client.start('hello');
+  var outgoing = client.Channel();
+  sla(client, outgoing, function(){
+    console.log('done');
+  });
+}
--- a/app.js
+++ b/app.js
@@ -4,26 +4,36 @@
 var ari = require('ari-client');
 var util = require('util');
 var sla = require('./lib/sla.js');
+var Q = require('q');
 
-ari.connect('http://127.0.0.1:8088', 'user', 'pass', clientLoaded);
+var connect = Q.denodeify(ari.connect);
+
+connect('http://127.0.0.1:8088', 'user', 'pass')
+  .then(function (client) {
+    clientLoaded(client);
+  })
+  .catch(errHandler);
 
 /**
  * Starts Stasis app 'sla' and initiates SLA application after ARI connection.
- * @param {Object} err - Error message.
  * @param {Object} client - Object that contains information from the ARI 
  *   connection.
  */
-function clientLoaded (err, client) {
-  if (err) {
-    throw err;
-  }
+function clientLoaded (client) {
   client.start('sla');
-  sla(client, function(err) {
-    if (err !== null) {
-      throw err;
-    } else {
-      console.log('Application Completed');
+  sla(client)
+    .then(console.log)
+    .catch(errHandler)
+    .finally(function () {
       process.exit(0);
-    }
-  });
+    });
+}
+
+/**
+ * Handles errors found in application.
+ * @param {Object} err - error from application.
+ */
+function errHandler(err) {
+  console.error(err);
+  throw err;
 }
--- a/lib/sla.js
+++ b/lib/sla.js
@@ -4,12 +4,12 @@
 /**
  * Receives initial input and begins the application.
  * @param {Object} client - Client received from app.js.
- * @callback stop - Ends the application.
+ * @callback end - Handles errors and ends the application.
  */
-module.exports = function(client, stop) {
+module.exports = function(client, end) {
   designateMixingBridge();
 
-  /*
+  /**
    * Looks for an existing mixing bridge and creates one if none exist.
    */
   function designateMixingBridge() {
@@ -19,18 +19,16 @@ module.exports = function(client, stop) {
       })[0];
       if (bridge) {
         console.log('Using existing mixing bridge %s', bridge.id);
-        createChannel(client, bridge);
       } else {
         bridge = client.bridges.create({type: 'mixing'}, function(err, bridge) {
           if (err) {
-            console.error(err);
-            process.exit(1);
+            end(err);
           } else {
             console.log('Created new mixing bridge %s', bridge.id);
-            createChannel(client, bridge);
           }
         });
       }
+      createChannel(client, bridge);
     });
   }
 
@@ -43,9 +41,8 @@ module.exports = function(client, stop) {
     console.log('Adding channel to bridge %s', bridge.id);
     bridge.addChannel({channel: channel.id}, function(err) {
       if (err) {
-        console.error(err);
         channel.hangup();
-        process.exit(1);
+        end(err);
       }
     });
     bridge.on('ChannelEnteredBridge', function(event) {
@@ -53,7 +50,7 @@ module.exports = function(client, stop) {
     });
     bridge.on('ChannelLeftBridge', function(event) {
       console.log("Channel %s has left the bridge", event.channel.id);
-      stop();
+      end(null);
     });
   }
 
@@ -69,8 +66,7 @@ module.exports = function(client, stop) {
     channel.originate({endpoint: 'SIP/phone', app: 'sla'},
       function(err) {
         if (err) {
-          console.error(err);
-          process.exit(1);
+          end(err);
         }
       }
     );
--- a/lib/sla.js
+++ b/lib/sla.js
@@ -1,77 +1,77 @@
 /*jslint node: true */
-'use strict';
+'use strict'
 
 /**
- * Receives initial input and begins the application.
- * @param {Object} client - Client received from app.js.
- * @callback end - Handles errors and ends the application.
+ * Looks for an existing mixing bridge and creates one if none exist.
  */
-module.exports = function(client, end) {
-  designateMixingBridge();
-
-  /**
-   * Looks for an existing mixing bridge and creates one if none exist.
-   */
-  function designateMixingBridge() {
-    client.bridges.list(function(err, bridges) {
-      var bridge = bridges.filter(function(candidate) {
-        return candidate.bridge_type === 'mixing';
-      })[0];
-      if (bridge) {
-        console.log('Using existing mixing bridge %s', bridge.id);
-      } else {
-        bridge = client.bridges.create({type: 'mixing'}, function(err, bridge) {
-          if (err) {
-            end(err);
-          } else {
-            console.log('Created new mixing bridge %s', bridge.id);
-          }
-        });
-      }
-      createChannel(client, bridge);
-    });
-  }
-
-  /**
-   * Adds created channel to the bridge.
-   * @param {Object} channel - Created channel to be added to the bridge.
-   * @param {Object} bridge - Bridge that channel is to be added to.
-   */
-  function addChannelsToBridge(channel, bridge) {
-    console.log('Adding channel to bridge %s', bridge.id);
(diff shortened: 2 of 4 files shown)


app.js
```diff
@@ -1,10 +1,27 @@
+/*jslint node: true */
```
samuelg, line 1: Look at adding a .jshintrc file to drive this type of jshint configuration. See https://raw.githubusercontent.com/asterisk/node-voicemail-data/master/.jshintrc for an example.

Update: You no longer need /*jslint node: true */ , your .jshintrc file is taking care of that.

app.js
```diff
@@ -1,10 +1,27 @@
+/*jslint node: true */  
 'use strict';
 
 var ari = require('ari-client');
 var util = require('util');
 var sla = require('./lib/sla.js');
 
-ari.connect('http://127.0.0.1:8088', 'ariUser', 'boogers', clientLoaded);
+ari.connect('http://127.0.0.1:8088', 'user', 'pass', clientLoaded);
```
samuelg: You will eventually want to move the URL/user/pass of the connect method into a config file loaded by this file.

app.js
```diff
@@ -1,10 +1,27 @@
...
 
-function clientLoaded (err, client){
-    console.log(sla());}
+/**
+ * Starts Stasis app 'sla' and initiates SLA application after ARI connection.
+ * @param {Object} err - Error message.
+ * @param {Object} client - Object that contains information from the ARI 
+ *   connection.
+ */
+function clientLoaded (err, client) {
```
samuelg: You never check to see if the connection returned an error.

Update: looks good.

app.js
```diff
@@ -1,10 +1,27 @@
...
+  client.start('sla');
+  sla(client, function(err) {
+    if (err != null) {
```
samuelg: You can just test for if (err)

lib/sla.js
```diff
@@ -1,6 +1,77 @@
...
-module.exports = function(callback){
-    console.log("\"snarflegluff\" should be printed below:");
-    var str = "snarflegluff";
-    return str;}
+/**
+ * Receives initial input and begins the application.
+ * @param {Object} client - Client received from app.js.
+ * @callback end - Handles errors and ends the application.
+ */
+module.exports = function(client, end) {
```
samuelg: It's common to use the name callback for callback functions as it signals an asynchronous API.

lib/sla.js
```diff
@@ -1,6 +1,77 @@
...
+  designateMixingBridge();
+
+  /**
+   * Looks for an existing mixing bridge and creates one if none exist.
+   */
+  function designateMixingBridge() {
```
samuelg: Because you are defining these functions inside the module.exports function, they will be created every time require is called on this file. You should move them above the export statement so they only need to be created once.

lib/sla.js
```diff
@@ -1,6 +1,77 @@
...
+    client.bridges.list(function(err, bridges) {
+      var bridge = bridges.filter(function(candidate) {
+        return candidate.bridge_type === 'mixing';
+      })[0];
+      if (bridge) {
+        console.log('Using existing mixing bridge %s', bridge.id);
+      } else {
+        bridge = client.bridges.create({type: 'mixing'}, function(err, bridge) {
```
samuelg: This is an asynchronous call, you can't assign the result of it to bridge. The bridge object that is passed into the callback is the object you are looking for.

lib/sla.js
```diff
@@ -1,6 +1,77 @@
...
+  /**
+   * Creates a channel to a specified endpoint and places it in the Stasis app.
+   * @param {Object} client - Object that contains information from the ARI
+   *   connection.
+   * @param {Object} bridge - Object that defines the bridge to be passed to
+   *   addChannelsToBridge().
+   */
+  function createChannel(client, bridge) {
+    var channel = client.Channel();
+    channel.originate({endpoint: 'SIP/phone', app: 'sla'},
```
samuelg: The endpoint parameter should probably be passed into the module instead of being hard-coded.

lib/sla.js
```diff
@@ -1,6 +1,77 @@
...
+      function(err) {
+        if (err) {
+          end(err);
+        }
+      }
+    );
+    channel.on('StasisStart', function(event, outgoing) {
```
samuelg: outgoing seems vague in this case. I would consider changing the name to something a bit more descriptive.

package.json
```diff
@@ -3,46 +3,28 @@
   "version": "1.0.0",
```
samuelg: Read http://semver.org/. Typically, the first version of a module is 0.0.1. Once the first release is out, this changes to 0.1.0. After that, every time a feature is added, the middle number (minor version) is incremented. Every time a bug is fixed, the last number (patch version) is incremented. Every time a backwards incompatible change is added, the first number (major version) is incremented. Some modules are go to 1.0.0 once the API is finalized and the module is stable.

package.json
```diff
@@ -3,46 +3,28 @@
...
   "description": "SLA Application using node-ari-client.js",
```
samuelg: As far as NPM is concerned, the client is ari-client. node-ari-client is only the name on github.

README.md
```diff
@@ -1,2 +1,4 @@
 # node-ari-shared-line-appearances
 Co-op implementation of SLA in ARI using Node.js
+
+This implementation assumes that you have at least Asterisk 12.0.0 running. This is the version where ARI first came around, and it should have all of the functions required for this project. You must also have a valid ARI user in ari.conf named user and have a password "pass", as well as have 127.0.0.1 (or localhost) and 8088 configured as the bindaddr and bindport respectively in http.conf. You must have a SIP user named phone, or change the originate to match a valid SIP device.
```
samuelg: Things like the ARI user/pass and the SIP endpoint (or other technologies such as PJSIP) should not be given here. These values should come from a configuration file.

app.js
```diff
@@ -1,10 +1,27 @@
...
+      console.error(err);
+      process.exit(1);
+    } else {
+      console.log('Application Completed');
+      process.exit(0);
+    }
+  });
+};
```
samuelg: No need for a semi-colon here since you are using the function name() {} syntax.

app.js
samuelg: BTW, if you needed to check against null, you should use !== instead of !=.

samuelg:
Overall, instead of having a function that calls another function, which in turn calls another function and so on, consider having small functions responsible for a single thing that expect a callback so you can get the value you need (the bridge to add to, the channel object) and having the main export function call these in turn. This will make the module more readable and will help in maintenance down the line.

lib/sla.js
samuelg: That particular channel is only ever going to enter into Stasis once. Use channel.once('StasisStart') instead of .on so the event handler can unregister itself once the event is raised and the handler can be garbage collected.

lib/sla.js
```diff
@@ -1,6 +1,77 @@
...
+   */
+  function addChannelsToBridge(channel, bridge) {
+    console.log('Adding channel to bridge %s', bridge.id);
+    bridge.addChannel({channel: channel.id}, function(err) {
+      if (err) {
+        channel.hangup();
+        end(err);
+      }
+    });
+    bridge.on('ChannelEnteredBridge', function(event) {
```
samuelg: That particular channel is only ever going to enter the bridge once. Use channel.once instead of .on so the event handler can unregister itself once the event is raised and the handler can be garbage collected.

lib/sla.js
```diff
@@ -1,6 +1,77 @@
...
+      console.log("Channel %s has entered the bridge", event.channel.id);
+    });
+    bridge.on('ChannelLeftBridge', function(event) {
```
samuelg: That particular channel is only ever going to leave the bridge once. Use channel.once instead of .on so the event handler can unregister itself once the event is raised and the handler can be garbage collected.

app.js
TACambron: That is coming up in one of the next assignments.

lib/sla.js
wolfeman67: Configuration file will come later :)

package.json
wolfeman67: Version number was fixed in one of the bridges, before being merged to master. An incorrect version number was used in master previously. But that has been corrected now.

lib/sla.js
TACambron: Just for this assignment, Matt wanted us to hard code an endpoint. We will adjust this in later versions.

lib/sla.js
```diff
@@ -1,6 +1,77 @@
...
+function createChannel(client, bridge, callback) {
+  var channel = client.Channel();
+  channel.originate({endpoint: 'SIP/phone', app: 'sla'},
+    function(err) {
+      if (err) {
+        callback(err);
+      }
+    }
+  );
+  channel.once('StasisStart', function(event) {
```
samuelg: this should read: channel.once('StasisStart', function(event, channel) {

lib/sla.js
```diff
@@ -1,6 +1,77 @@
...
+    addChannelsToBridge(event.channel, bridge, callback);
```
samuelg: use the channel object passed in as the second parameter for this. event.channel is a simple object literal that contains data only. The object passed in to the second parameter is an object with methods for further operations and responds to event listeners.

app.js
samuelg: You no longer need /*jslint node: true */ , your .jshintrc file is taking care of that.

app.js
```diff
@@ -6,5 +7,23 @@ var sla = require('./lib/sla.js');
...
+ */
+function clientLoaded (err, client) {
+  if (err) {
+    throw err;
+  }
+  client.start('sla');
+  sla(client, function(err) {
+    if (err !== null) {
+      throw err;
+    } else {
```
samuelg: No need for the else, throw will kill the process there.

lib/sla.js
```diff
@@ -1,6 +1,77 @@
-'use strict';
+/*jslint node: true */
```
samuelg: This should be removed now that you have a .jshintrc file.

lib/sla.js
```diff
@@ -1,6 +1,77 @@
...
+'use strict'
 
-module.exports = function(callback){
-    console.log("\"snarflegluff\" should be printed below:");
-    var str = "snarflegluff";
-    return str;}
+/**
+ * Looks for an existing mixing bridge and creates one if none exist.
```
samuelg, line 5: jsdoc style param list is missing.

lib/sla.js
```diff
@@ -1,6 +1,77 @@
...
+ */
+function designateMixingBridge(client, callback) {
+  client.bridges.list(function(err, bridges) {
+    var bridge = bridges.filter(function(candidate) {
+      return candidate.bridge_type === 'mixing';
```
samuelg: Is this the expected behavior? You never check to see if the bridge has channel(s) in it and since it's a mixing bridge, the new channel will hear whatever other channels are currently hearing and vice versa.

lib/sla.js
```diff
@@ -1,6 +1,77 @@
...
+    })[0];
+    if (bridge) {
+      console.log('Using existing mixing bridge %s', bridge.id);
+      createChannel(client, bridge, callback);
+    } else {
+      client.bridges.create({type: 'mixing'}, function(err, bridge) {
+        if (err) {
+          callback(err);
```
samuelg: Just as an FYI, calling a callback does not exit the function at this point. Since you are not returning from the function, the call to createChannel will occur even though an error. This means that createChannel will eventually error out and call the callback yet again, which is very bad. This is another thing that is bad about callbacks; even though they should only be called once, it's very easy to accidentally call them multiple times.

lib/sla.js
```diff
@@ -1,6 +1,77 @@
...
+ * Adds created channel to the bridge.
+ * @param {Object} channel - Created channel to be added to the bridge.
+ * @param {Object} bridge - Bridge that channel is to be added to.
+ */
+function addChannelsToBridge(channel, bridge, callback) {
+  console.log('Adding channel to bridge %s', bridge.id);
+  bridge.addChannel({channel: channel.id}, function(err) {
+    if (err) {
+      channel.hangup();
+      callback(err);
```
samuelg: See comment above, this will not actually exit the function and the event handlers will be registered even in the case of an error occurring.

lib/sla.js
```diff
@@ -1,6 +1,77 @@
...
+ *   connection.
+ * @param {Object} bridge - Object that defines the bridge to be passed to
+ *   addChannelsToBridge().
+ */
+function createChannel(client, bridge, callback) {
+  var channel = client.Channel();
+  channel.originate({endpoint: 'SIP/tcam', app: 'sla'},
+    function(err) {
+      if (err) {
+        callback(err);
```
samuelg: See comment above, this will not actually exit the function and the event handler will be registered even in the case of an error occurring.

samuelg:
At this point, it would be a good idea to switch from callbacks to promises. You should use promises both internally to handle calling async APIs and also return a promise to users of your module that will eventually be resolved if everything works as expected, or rejected if an error occurs. You should take some time to read about promises and pick a library that implements them in Node.js.

Each of your method should have a single purpose, and should return a promise that can be used by the caller to determine if the module should be rejected, or if the next method should be called to handle the next part of the module's responsibilities.

In other words, the main function exposed as your module should look something like the following (note that I'm just reusing your current method names as an example, you may want to design the API differently):

return designateMixingBridge(args)
  .then(function(result) {
    // do one of these
    return createBridge(args);
    return createChannel(args);
  })
  .then(function(result) {
    return addChannelToBridge(args);
  });

Note that .done is not called on the promise since it is returned to the user of the module.

You will need to look into converting the Node.js style callback API of ari-client to a promise one using something like Q's denodeify method so you can call ari-client methods and get back promises.

samuelg:
Running jshint found the following errors:

lib/sla.js: line 2, col 1, Missing semicolon.
lib/sla.js: line 10, col 0, Identifier 'bridge_type' is not in camel case.
lib/sla.js: line 41, col 52, Mixed double and single quotes.
lib/sla.js: line 44, col 49, Mixed double and single quotes.
lib/sla.js: line 71, col 81, Line is too long.
lib/sla.js: line 72, col 81, Line is too long.
lib/sla.js: line 73, col 86, Line is too long.
lib/sla.js: line 75, col 86, Line is too long.

lib/sla.js
matt-jordan: For this current task, yes, that's the expected behaviour. Later on, they'll have to be a bit smarter about how they select the bridges.

app.js
samuelg, line 1: /*jslint node: true */ is no longer necessary.

app.js
```diff
@@ -1,10 +1,39 @@
...
-ari.connect('http://127.0.0.1:8088', 'user', 'pass', clientLoaded);
+var connect = Q.denodeify(ari.connect);
 
-function clientLoaded (err, client){
-    console.log(sla());}
+connect('http://127.0.0.1:8088', 'user', 'pass')
+  .then(function (client) {
+    clientLoaded(client);
+  })
+  .catch(errHandler);
```
samuelg: You should call done here so the error you throw in errHandler can propagate up the stack.

app.js
```diff
@@ -1,10 +1,39 @@
...
 var sla = require('./lib/sla.js');
+var Q = require('q');
 
-ari.connect('http://127.0.0.1:8088', 'user', 'pass', clientLoaded);
+var connect = Q.denodeify(ari.connect);
 
-function clientLoaded (err, client){
-    console.log(sla());}
+connect('http://127.0.0.1:8088', 'user', 'pass')
+  .then(function (client) {
```
samuelg: you can just use .then(clientLoaded)

app.js
```diff
@@ -1,10 +1,39 @@
...
+ * @param {Object} client - Object that contains information from the ARI 
+ *   connection.
+ */
+function clientLoaded (client) {
+  client.start('sla');
+  sla(client)
+    .then(console.log)
+    .catch(errHandler)
+    .finally(function () {
+      process.exit(0);
```
samuelg: Since this is in a finally handler, process will exit with a non error 0 code even in the case of an error.

app.js
```diff
@@ -1,10 +1,39 @@
...
+ * Starts Stasis app 'sla' and initiates SLA application after ARI connection.
+ * @param {Object} client - Object that contains information from the ARI 
+ *   connection.
+ */
+function clientLoaded (client) {
+  client.start('sla');
+  sla(client)
+    .then(console.log)
+    .catch(errHandler)
+    .finally(function () {
```
samuelg: You should call done as the final call to the promise.

lib/sla.js
```diff
@@ -1,6 +1,93 @@
...
-    return str;}
+/**
+ * Looks for an existing mixing bridge and creates one if none exist.
+ * @param {Object} client - Object that contains information from the ARI
+ *   connection.
+ * @return {Q} Q promise object.
+ */
+function designateMixingBridge(client) {
+  var defer = Q.defer();
+  var list = Q.denodeify(client.bridges.list.bind(client.bridges));
```
samuelg: You actually want to bind to client here, not client.bridges. Any time you denodeify a function called on client.someResource you pass the client to bind. Any time you denodeify a function called on a resource instance like bridge.get you pass the bridge to bind.

lib/sla.js
```diff
@@ -1,6 +1,93 @@
...
+  list().then(
+      function(bridges) {
+        var bridge = bridges.filter(function(candidate) {
+          return candidate['bridge_type'] === 'mixing';
+        })[0];
+        if (bridge) {
+          console.log('Using existing mixing bridge %s', bridge.id);
+          defer.resolve(bridge);
+        } else {
+          client.bridges.create({type: 'mixing'}, function(err, bridge) {
```
samuelg: This is still using the callback version of bridges.create and should be switched to promise style.

lib/sla.js
```diff
@@ -1,6 +1,93 @@
...
+  addChannel({channel: channel.id})
+    .catch(function (err) {
+      defer.reject(err);
+    });
+  bridge.once('ChannelEnteredBridge', function(event, bridge) {
+    console.log('Channel %s has entered the bridge', bridge.channel.id);
+  });
+  bridge.once('ChannelLeftBridge', function(event, bridge) {
+    console.log('Channel %s has left the bridge', bridge.channel.id);
+    defer.resolve('Application completed');
```
samuelg, line 55: That's fine for this first step, but I don't think you want to ultimately have the promise resolve only once the channel hangs up. Events are a better way of informing users of your API that something like that has happened.

lib/sla.js
```diff
@@ -1,6 +1,93 @@
...
+
+/**
+ * Receives initial input and begins the application.
+ * @param {Object} client - Client received from app.js.
+ * @return {Q} Q promise object.
+ */
+module.exports = function(client) {
+  return designateMixingBridge(client)
+    .then(function (bridge) {
+      return createChannel(client, bridge)
```
samuelg, line 90: This is fine for this first step but there's probably a way in the future to make this flatter and not nest the calls like this.

samuelg:
Looks good. Since you have a lot of commits to address PR comments, I would suggest we walk though squashing the commits into a single commit in a different branch before merging.

TACambron:
Squashed into #6

PR #5, opened 2015-02-10
Added unit tests and a Gruntfile

wolfeman67:
The Gruntfile runs jshint and the test file that is located in the tests folder.
The tests include checking:
1. If a bridge is created when there are none to be found.
2. If a bridge is used again if a preexisting one is found.
3. If a channel is allowed to continue (from originating) when a valid endpoint is given.
4. If a channel isn't allowed to continue (from originating) when an invalid endpoint is given.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,20 +1,21 @@
 module.exports = function(grunt) {
-   grunt.initConfig({
-       jshint: {
-          files: ['Gruntfile.js', 'test/**/*.js', 'lib/*.js', '*.js']
+  grunt.initConfig({
+    jshint: {
+      files: ['Gruntfile.js', 'tests/**/*.js', 'lib/**/*.js', '*.js']
+    },
+    mochaTest: {
+      test: {
+        options: {
+          mocha: require('mocha'),
+          reporter: 'spec',
+          timeout: 2000
         },
-       mochaTest: {
-         test: {
-           options: {
-             mocha: require('mocha'),
-             reporter: 'spec',
-             timeout: 2000
-           },
-           src: ['tests/*.js']
-         }
-       }});
-    grunt.loadNpmTasks('grunt-contrib-jshint');
-    grunt.loadNpmTasks('grunt-mocha-test');
+        src: ['tests/*.js']
+      }
+    }
+  });
+  grunt.loadNpmTasks('grunt-contrib-jshint');
+  grunt.loadNpmTasks('grunt-mocha-test');
 
-    grunt.registerTask('default', ['jshint', 'mochaTest']);
-    };
+  grunt.registerTask('default', ['jshint', 'mochaTest']);
+};
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,0 +1,20 @@
+module.exports = function(grunt) {
+   grunt.initConfig({
+       jshint: {
+          files: ['Gruntfile.js', 'test/**/*.js', 'lib/*.js', '*.js']
+        },
+       mochaTest: {
+         test: {
+           options: {
+             mocha: require('mocha'),
+             reporter: 'spec',
+             timeout: 2000
+           },
+           src: ['tests/*.js']
+         }
+       }});
+    grunt.loadNpmTasks('grunt-contrib-jshint');
+    grunt.loadNpmTasks('grunt-mocha-test');
+
+    grunt.registerTask('default', ['jshint', 'mochaTest']);
+    };
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,6 +1,9 @@
 module.exports = function(grunt) {
   grunt.initConfig({
     jshint: {
+      options: {
+        jshintrc: true
+      },
       files: ['Gruntfile.js', 'tests/**/*.js', 'lib/**/*.js', '*.js']
     },
     mochaTest: {
--- a/tests/sample.js
+++ b/tests/sample.js
@@ -1,0 +1,199 @@
+/**
+ * SLA bridge and channel creation tests.
+ *
+ */
+"use strict";
+
+/*global describe:false*/
+/*global afterEach:false*/
+
+var util = require('util');
+var Emitter = require('events').EventEmitter;
+
+var mockClient;
+var mockChannel;
+var mockBridge;
+// The channels in existance
+var channels = [];
+// The bridges in existance (should only be one)
+var bridges = [];
+// The channels in the bridges
+var bridge_channels = [];
+// Tests if a bridge has been created or not on a particular test. Resets
+var bridge_created = 0;
+// Mock value for whether the endpoint is valid or not
+var valid = 1;
+// Conditional for whether or not StasisStart has been passed yet
+var pastStasis = 0;
+
+// The mocked up version of the callback error function
+var end= function(err){console.log(err);}
+
+// Millesecond delay for mock requests
+var asyncDelay = 100;
+
+/**
+ * Returns a mock client that houses the bridges and can create both bridges
+ * and channels.
+ */
+
+var getMockClient = function() {
+  if(mockClient) {
+    return mockClient;
+  }
+  var Client = function() {
+    this.bridges = {
+      list: function(cb) {
+        console.log(bridges);
+        cb(null, bridges);
+        return bridges;
+      },
+      create: function(param, cb) {
+        bridge_created = 1;
+        console.log(param);
+        bridges.push(getMockBridge(param));
+        console.log(bridges)
+        cb(null, bridges[0]);
+      }
+    };
+    this.Channel= function() {
+      return getMockChannel();
+      }
+  };
+  mockClient = new Client();
+  return mockClient;
+};
+
+/**
+ * Returns a mock bridge that has a bridge_type and id, and can add channels
+ * to its array of channels. Emits a ChannelEnteredBridge event
+ */
+
+var getMockBridge = function(param) {
+  var Bridge = function(param) {
+    console.log(param.type);
+    this.bridge_type = param.type;
+    this.id = "1";
+    this.addChannel= function(input, cb) {
+      console.log("TREX");
+      pastStasis = 1;
+      var channel = channels.filter(function(testChan) {
+        return testChan.id === '1';
+      })[0];
+      bridge_channels.push(channel);
+      cb(null);
+      console.log(channel);
+      this.emit('ChannelEnteredBridge', {channel: {id: channel.id}});
+    };
+  };
+  util.inherits(Bridge, Emitter);
+  mockBridge = new Bridge(param);
+  return mockBridge;
+};
+
+/**
+ * Returns a mock channel that has an id, and can use originate to add a channel
+ * to the array of total channels. Emits a StasisStart event when originating.
+ */
+
+var getMockChannel = function() {
+  var Channel = function() {
+    this.id = '1';
+    this.originate = function(input, cb) {
+      console.log(input);
+      var self = this;
+      setTimeout(function() {
+        if(valid == 1) {
+          console.log(self);
+          self.emit('StasisStart', self, self);
+        }
+      }, 20);
+    };
+  };
+  util.inherits(Channel, Emitter);
+  mockChannel = new Channel();
+  channels.push(mockChannel);
+  return mockChannel;
+};
+
+describe('SLA Bridge and Channel Tester', function() {
+
+  afterEach(function(done) {
+    channels = [];
+    bridge_channels = [];
+    bridge_created = 0;
+    done();
+    });
+
+  it('should create a bridge when there isn\'t one', function(done) {
+    var channel = getMockChannel();
+    var client = getMockClient();
+    var sla = require('../lib/sla.js')(client, end);
+
+    bridgeChecking();
+    function bridgeChecking() {
+      setTimeout(function() {
+        if(channels.length != 0 && bridge_created == 1) {
+          done();
+        } else {
+          bridgeChecking();
+        }
+      }, asyncDelay);
+    } 
+  });
+
+  it('should use a preexisting bridge if there is one', function(done) {
+    var channel = getMockChannel();
+    var client = getMockClient();
+    var sla = require('../lib/sla.js')(client, end);
+
+    bridgeChecking();
+    function bridgeChecking() {
+      setTimeout(function() {
+        if(channels.length != 0 && bridge_created == 0) {
+          done();
+        } else {
+          bridgeChecking();
+        }
+      }, asyncDelay);
+    } 
+  });
+
+  it('should use a "valid" endpoint and give off a StasisStart event ' +
+     'to proceed to the next section', function(done) {
+    var channel = getMockChannel();
+    var client = getMockClient();
+    var sla = require('../lib/sla.js')(client, end);
+
+    validChecking();
+    function validChecking() {
+      setTimeout(function() {
+        if(pastStasis == 1 && bridge_channels.length != 0) {
+          pastStasis = 0;
+          done();
+        } else {
+          validChecking();
+        }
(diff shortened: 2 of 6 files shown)


Gruntfile.js
```diff
@@ -0,0 +1,20 @@
+module.exports = function(grunt) {
+   grunt.initConfig({
+       jshint: {
+          files: ['Gruntfile.js', 'test/**/*.js', 'lib/*.js', '*.js']
```
samuelg: If you use the /*_/_.js syntax you should also apply it to lib/. Instead of specifying which subdirectories you care about, you are saying to check all subdirectories under test/ or lib/.

Gruntfile.js
```diff
@@ -0,0 +1,20 @@
...
+             timeout: 2000
+           },
+           src: ['tests/*.js']
+         }
+       }});
+    grunt.loadNpmTasks('grunt-contrib-jshint');
+    grunt.loadNpmTasks('grunt-mocha-test');
+
+    grunt.registerTask('default', ['jshint', 'mochaTest']);
+    };
```
samuelg: Several lines here are not aligned correctly and makes the file hard to read. This close }; should match the indentation of the module.exports line. The grunt.loadNpmTasks lines should match the indentation of the grunt.initConfig line. The closing }}); before the first grunt.loadNpmTests line should match the indentation of the grunt.initConfig line as well.

Gruntfile.js
```diff
@@ -0,0 +1,20 @@
+module.exports = function(grunt) {
+   grunt.initConfig({
+       jshint: {
```
samuelg: You have jshint configured but you have no .jshintrc file to drive the behavior, see https://raw.githubusercontent.com/asterisk/node-voicemail-data/master/.jshintrc for an example of how to configure jshint.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+var bridge_channels = [];
+// Tests if a bridge has been created or not on a particular test. Resets
+var bridge_created = 0;
+// Mock value for whether the endpoint is valid or not
+var valid = 1;
+// Conditional for whether or not StasisStart has been passed yet
+var pastStasis = 0;
+
+// The mocked up version of the callback error function
+var end= function(err){console.log(err);}
```
samuelg: should be a space after end and after the closing ) of the function signature. I would also move console.log on the next line.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+};
+
+describe('SLA Bridge and Channel Tester', function() {
+
+  afterEach(function(done) {
+    channels = [];
+    bridge_channels = [];
+    bridge_created = 0;
+    done();
+    });
```
samuelg: This line should match the indentation of the afterEach line.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+
+  it('should create a bridge when there isn\'t one', function(done) {
+    var channel = getMockChannel();
+    var client = getMockClient();
+    var sla = require('../lib/sla.js')(client, end);
```
samuelg: Calling the sla module this way with an error callback that simply logs the error means you can't test the module for failure. You should be able to test for a failure in the sla module using different mocks or mocks configured diffently.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+  util.inherits(Channel, Emitter);
+  mockChannel = new Channel();
+  channels.push(mockChannel);
+  return mockChannel;
+};
+
+describe('SLA Bridge and Channel Tester', function() {
+
+  afterEach(function(done) {
+    channels = [];
```
samuelg: You should also reset bridges, valid, and pastStasis.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+"use strict";
+
+/*global describe:false*/
+/*global afterEach:false*/
+
+var util = require('util');
+var Emitter = require('events').EventEmitter;
+
+var mockClient;
+var mockChannel;
```
samuelg: I don't think you need this since you are storing mock channels for a given test in the channels array.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+var mockBridge;
```
samuelg: I don't think you need this since you are storing mock bridges for a given test in the bridges array.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+  }
+  var Client = function() {
+    this.bridges = {
+      list: function(cb) {
+        console.log(bridges);
+        cb(null, bridges);
+        return bridges;
+      },
+      create: function(param, cb) {
+        bridge_created = 1;
```
samuelg: Couldn't you just check the bridges array to see if a new bridge was created?

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+        console.log(param);
+        bridges.push(getMockBridge(param));
+        console.log(bridges)
+        cb(null, bridges[0]);
```
samuelg: This will always pass the first bridge that was created instead of the latest. Use bridges[bridge.length - 1] instead.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+      }
+    };
+    this.Channel= function() {
+      return getMockChannel();
+      }
```
samuelg: This should be aligned with the this.Channel line.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+  };
+  mockClient = new Client();
```
samuelg: You should be making this an EventEmitter so that you can emit events to the sla module.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+/**
+ * Returns a mock bridge that has a bridge_type and id, and can add channels
+ * to its array of channels. Emits a ChannelEnteredBridge event
+ */
+
+var getMockBridge = function(param) {
+  var Bridge = function(param) {
+    console.log(param.type);
+    this.bridge_type = param.type;
+    this.id = "1";
```
samuelg: You should increment ids so more than one bridge can be created with unique IDs.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+    this.addChannel= function(input, cb) {
+      console.log("TREX");
```
samuelg: console.log statements are ok while you are writing tests but generally do not get included as part of the tests. Tests report whether they pass or not to the terminal but usually no more.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+      pastStasis = 1;
```
samuelg: I'm not clear what this is for?

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+      var channel = channels.filter(function(testChan) {
+        return testChan.id === '1';
```
samuelg: the channel will be passed to the method addChannel, you shoudn't get a channel this way to "add" it to the bridge. In general, your mock should work like the real thing. If it expects a channel in the arguments, you should use that.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+      })[0];
+      bridge_channels.push(channel);
+      cb(null);
+      console.log(channel);
+      this.emit('ChannelEnteredBridge', {channel: {id: channel.id}});
```
samuelg: the event object passed to the event should also include info on the bridge. Additionally, a second object should be passed that include a channel and a bridge key with mock objects as values.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+var getMockChannel = function() {
+  var Channel = function() {
+    this.id = '1';
+    this.originate = function(input, cb) {
+      console.log(input);
+      var self = this;
+      setTimeout(function() {
+        if(valid == 1) {
+          console.log(self);
+          self.emit('StasisStart', self, self);
```
samuelg: The first object passed to the event should not simply be a channel but rather an event object that includes a channel key with data about the channel. If a library you are testing looks at event.channel.id for example it would fail.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+};
+
+/**
+ * Returns a mock channel that has an id, and can use originate to add a channel
+ * to the array of total channels. Emits a StasisStart event when originating.
+ */
+
+var getMockChannel = function() {
+  var Channel = function() {
+    this.id = '1';
```
samuelg: Use incrementing ids so you can create more than 1 mock channel with unique IDs.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+    this.originate = function(input, cb) {
+      console.log(input);
+      var self = this;
+      setTimeout(function() {
+        if(valid == 1) {
+          console.log(self);
+          self.emit('StasisStart', self, self);
+        }
+      }, 20);
```
samuelg: You should use an async delay variable instead of hardcoding this.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+describe('SLA Bridge and Channel Tester', function() {
+
+  afterEach(function(done) {
+    channels = [];
+    bridge_channels = [];
+    bridge_created = 0;
+    done();
+    });
+
+  it('should create a bridge when there isn\'t one', function(done) {
```
samuelg: Since sla.js is supposed to create a mixing bridge, you should validate that this is what occurs.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+        if(channels.length != 0 && bridge_created == 1) {
+          done();
+        } else {
+          bridgeChecking();
+        }
+      }, asyncDelay);
+    } 
+  });
+
+  it('should use a preexisting bridge if there is one', function(done) {
```
samuelg: This test appears to depend on the previous one running first, this may lead to problems in the future. Tests should be independent of each other. Create a bridge first and then import the sla module to test this. afterEach should reset all state related vars in this file.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+    var channel = getMockChannel();
+    var client = getMockClient();
+    var sla = require('../lib/sla.js')(client, end);
+
+    bridgeChecking();
+    function bridgeChecking() {
+      setTimeout(function() {
+        if(channels.length != 0 && bridge_created == 0) {
```
samuelg: This could pass even though the bridge could be about to be created. You may want to keep track of the following function in sla.js being called before you check if a bridge was created.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+          done();
+        } else {
+          bridgeChecking();
+        }
+      }, asyncDelay);
+    } 
+  });
+
+  it('should use a "valid" endpoint and give off a StasisStart event ' +
```
samuelg: Since this first task involves hard coding endpoints, you may want to specifically check for the expected input.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
...
+          pastStasis = 0;
+          done();
+        } else {
+          validChecking();
+        }
+      }, asyncDelay);
+    } 
+  });
+
+  it('should use an "invalid" endpoint and not give off a StasisStart event',
```
samuelg: This test is no different from the previous one, it's only passing because the check occurs before the call updates the variables you are checking. Until sla.js is made to expect parameters to drive what endpoint is called you may not be able to test this.

samuelg:
You should have grunt and all grunt plugins in your dev-dependencies in package.json. Currently grunt does not run after a checkout and npm install.

samuelg:
running jshint on tests/sample.js found the following errors:

tests/sample.js: line 5, col 1, Use the function form of "use strict".
tests/sample.js: line 30, col 42, Missing semicolon.
tests/sample.js: line 55, col 29, Missing semicolon.
tests/sample.js: line 61, col 8, Missing semicolon.
tests/sample.js: line 136, col 28, Use '!==' to compare with '0'.
tests/sample.js: line 153, col 28, Use '!==' to compare with '0'.
tests/sample.js: line 153, col 51, Use '===' to compare with '0'.
tests/sample.js: line 171, col 54, Use '!==' to compare with '0'.
tests/sample.js: line 191, col 23, Use '===' to compare with '0'.
tests/sample.js: line 191, col 54, Use '===' to compare with '0'.
tests/sample.js: line 10, col 12, 'require' is not defined.
tests/sample.js: line 11, col 15, 'require' is not defined.
tests/sample.js: line 30, col 24, 'console' is not defined.
tests/sample.js: line 47, col 9, 'console' is not defined.
tests/sample.js: line 53, col 9, 'console' is not defined.
tests/sample.js: line 55, col 9, 'console' is not defined.
tests/sample.js: line 74, col 5, 'console' is not defined.
tests/sample.js: line 78, col 7, 'console' is not defined.
tests/sample.js: line 85, col 7, 'console' is not defined.
tests/sample.js: line 103, col 7, 'console' is not defined.
tests/sample.js: line 105, col 7, 'setTimeout' is not defined.
tests/sample.js: line 107, col 11, 'console' is not defined.
tests/sample.js: line 128, col 3, 'it' is not defined.
tests/sample.js: line 131, col 15, 'require' is not defined.
tests/sample.js: line 135, col 7, 'setTimeout' is not defined.
tests/sample.js: line 145, col 3, 'it' is not defined.
tests/sample.js: line 148, col 15, 'require' is not defined.
tests/sample.js: line 152, col 7, 'setTimeout' is not defined.
tests/sample.js: line 162, col 3, 'it' is not defined.
tests/sample.js: line 166, col 15, 'require' is not defined.
tests/sample.js: line 170, col 7, 'setTimeout' is not defined.
tests/sample.js: line 181, col 3, 'it' is not defined.
tests/sample.js: line 186, col 15, 'require' is not defined.
tests/sample.js: line 190, col 7, 'setTimeout' is not defined.

Some of these are due to your missing a .jshintrc file.

tests/sample.js
```diff
@@ -0,0 +1,199 @@
+/**
```
samuelg: Why is this file name sample.js? If this will be the only test file (small module testing), it should be called tests.js. If you will be testing multiple modules without the repo, it should probably be called sla.js.

tests/sample.js
wolfeman67: Will do later when this becomes necessary. For now, it is getting skipped.

tests/sla_tests.js
```diff
@@ -0,0 +1,212 @@
...
+  });
+
+  it('should create a bridge when there isn\'t one', function(done) {
+    var client = getMockClient();
+    var sla = require('../lib/sla.js')(client).done();
+
+    bridgeChecking();
+    function bridgeChecking() {
+      setTimeout(function() {
+        if(channels.length !== 0 && usingExisting === false &&
```
samuelg: since if statements already test for truth, you can just use  !usingExisting instead of usingExisting === false and isMixing instead of isMixing === true. Also, because 0 evaluates to false, you can just use channels.length instead of channels.length !== 0 but you can leave that if you want to make it explicit.

tests/sla_tests.js
```diff
@@ -0,0 +1,212 @@
...
+    this['bridge_type'] = param.type;
+    if (this['bridge_type'] === 'mixing') {
+      isMixing = true;
+    }
+    this.id = bridgeId.toString();
+    bridgeId += 1;
+    this.addChannel= function(input, cb) {
+      //Here channels and bridges should have the same ID number
+      var channel = channels.filter(function(testChan) {
+        return testChan.id === input.id;
```
samuelg: You have not addressed this yet. The input to addChannel looks like {channel: 'id'} and not {id: 'id'}.

samuelg:
Looks good. Since you have a lot of commits to address PR findings, I would suggest we walk though squashing the commits into a single commit in a different branch before merging.

wolfeman67:
Squashed into #7

PR #8, opened 2015-02-17
Made it possible for inbound callers to dial channels

wolfeman67:
If the SLA bridge (specified in the dialplan) is numeric,
then create or use a preexisting bridge that has the same
name as the value specified in the dialplan, create a new channel
to dial a specified endpoint.  If the endpoint hangs up, hang up
inbound caller. If the inbound caller hangs up, hang up dialed end.
Application resets when all have hung up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,3 +4,5 @@ Co-op implementation of SLA in ARI using Node.js
 This implementation assumes that you have at least Asterisk 12.0.0 running. This is the version where ARI first came around, and it should have all of the functions required for this project.
 
 You must also have a valid ARI user in ari.conf named "user" and have a password "pass", as well as have 127.0.0.1 (or localhost) and 8088 configured as the bindaddr and bindport respectively in http.conf
+
+You must also have a dialplan extension in extensions.conf that leads to the application (must have same name as what application is being started in the code) and that has an argument to represent the SLA bridge to reach.
--- a/README.md
+++ b/README.md
@@ -5,4 +5,10 @@ This implementation assumes that you have at least Asterisk 12.0.0 running. This
 
 You must also have a valid ARI user in ari.conf named "user" and have a password "pass", as well as have 127.0.0.1 (or localhost) and 8088 configured as the bindaddr and bindport respectively in http.conf
 
-You must also have a dialplan extension in extensions.conf that leads to the application (must have same name as what application is being started in the code) and that has an argument to represent the SLA bridge to reach.
+You must also have a dialplan extension in extensions.conf that leads to the application (must have same name as what application is being started in the code) and that has an argument to represent the SLA bridge to reach.  An example is below:
+
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+exten => 99,1,NoOp()                                                             
+    same => n,Stasis(sla,999)                                                    
+    same => n,Hangup()
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/app.js
+++ b/app.js
@@ -12,6 +12,10 @@ connect('http://127.0.0.1:8088', 'user', 'pass')
   .catch(errHandler)
   .done();
 
+/**
+ * Checks if the argument tied to the StasisStart event is dialed (not inbound)
+ * @param {String} argument - The argument (either an extension # or dialed)
+ */
 function isDialed(argument) {
   if(argument === 'dialed') {
     return true;
--- a/app.js
+++ b/app.js
@@ -1,4 +1,3 @@
-/*jslint node: true */  
 'use strict';
 
 var ari = require('ari-client');
@@ -14,26 +13,21 @@ connect('http://127.0.0.1:8088', 'user', 'pass')
   .done();
 
 /**
- * Starts Stasis app 'sla' and initiates SLA application after ARI connection.
+ * Waits for a StasisStart event before going into the main SLA module
  * @param {Object} client - Object that contains information from the ARI 
  *   connection.
  */
 function clientLoaded (client) {
   client.start('sla');
-  var defer = Q.defer();
   client.on('StasisStart', function(event, channel) {
-    stasisStart(event, client, channel);
+    if(event.args[0] !== 'dialed') {
+      var bridgeName = event.args[0];
+      sla(client, channel, bridgeName)
+        .then(console.log)
+        .catch(errHandler)
+        .done();
+    }
   });
-}
-
-function stasisStart (event, client, channel) {
-  if(event.args[0] !== 'dialed') {
-    var bridgeNumber = event.args[0];
-    sla(client, channel, bridgeNumber)
-      .then(console.log)
-      .catch(errHandler)
-      .done();
-  }
 }
 
 /**
--- a/app.js
+++ b/app.js
@@ -17,11 +17,7 @@ connect('http://127.0.0.1:8088', 'user', 'pass')
  * @param {String} argument - The argument (either an extension # or dialed)
  */
 function isDialed(argument) {
-  if(argument === 'dialed') {
-    return true;
-  } else {
-    return false;
-  }
+  return argument === 'dialed';
 }
 
 /**
--- a/app.js
+++ b/app.js
@@ -20,10 +20,20 @@ connect('http://127.0.0.1:8088', 'user', 'pass')
  */
 function clientLoaded (client) {
   client.start('sla');
-  sla(client)
-    .then(console.log)
-    .catch(errHandler)
-    .done();
+  var defer = Q.defer();
+  client.on('StasisStart', function(event, channel) {
+    stasisStart(event, client, channel);
+  });
+}
+
+function stasisStart (event, client, channel) {
+  if(event.args[0] !== 'dialed') {
+    var bridgeNumber = event.args[0];
+    sla(client, channel, bridgeNumber)
+      .then(console.log)
+      .catch(errHandler)
+      .done();
+  }
 }
 
 /**
@@ -31,6 +41,5 @@ function clientLoaded (client) {
  * @param {Object} err - error from application.
  */
 function errHandler(err) {
-  console.error(err);
   throw err;
 }
--- a/app.js
+++ b/app.js
@@ -12,6 +12,14 @@ connect('http://127.0.0.1:8088', 'user', 'pass')
   .catch(errHandler)
   .done();
 
+function isDialed(argument) {
+  if(argument === 'dialed') {
+    return true;
+  } else {
+    return false;
+  }
+}
+
 /**
  * Waits for a StasisStart event before going into the main SLA module
  * @param {Object} client - Object that contains information from the ARI 
@@ -20,7 +28,7 @@ connect('http://127.0.0.1:8088', 'user', 'pass')
 function clientLoaded (client) {
   client.start('sla');
   client.on('StasisStart', function(event, channel) {
-    if(event.args[0] !== 'dialed') {
+    if(!isDialed(event.args[0])) {
       var bridgeName = event.args[0];
       sla(client, channel, bridgeName)
         .then(console.log)
@@ -35,5 +43,9 @@ function clientLoaded (client) {
  * @param {Object} err - error from application.
  */
 function errHandler(err) {
-  throw err;
+  if(err.name === 'EarlyHangup' || err.name === 'HangupFailure') {
+   console.log(err.message);
+  } else {
+   throw err;
+  } 
 }
--- a/lib/sla.js
+++ b/lib/sla.js
@@ -15,7 +15,7 @@ function designateMixingBridge(client, bridgeName) {
       function(bridges) {
         var bridge = bridges.filter(function(candidate) {
           return (candidate['bridge_type'] === 'mixing' &&
-            candidate['name'] === bridgeName);
+            candidate.name === bridgeName);
         })[0];
         if (bridge) {
           console.log('Using existing mixing bridge %s numbered %s', 
@@ -126,7 +126,8 @@ function originateChannel(client, inbound, bridge) {
   return defer.promise;
 }
 
-/** Simple utility function for checking if the bridgeName is numerical
+/** 
+ * Simple utility function for checking if the bridgeName is numerical
  * @param {String} bridgeName - the name of the bridge to be created/used
  * @return {boolean} whether or not the bridgeName is numerical
  */
@@ -138,7 +139,8 @@ function checkIfNum (string) {
     return false;
   }
 }
-/** Represents an error that has both a name and message.
+/** 
+ * Represents an error that has both a name and message.
  * @param {String} name - the name/type of the error
  * @param {String} message - the corresponding message
  * Mainly used to avoid crashing the program, as it does with regular errors.
@@ -151,6 +153,8 @@ function CustomError(name, message) {
 /**
  * Receives initial input and begins the application.
  * @param {Object} client - Client received from app.js.
+ * @param {Object} inbound - the inbound channel
+ * @param {String} bridgeName - the name of the bridge to be used/created
  * @return {Q} Q promise object.
  */
 module.exports = function(client, inbound, bridgeName) {
@@ -170,7 +174,7 @@ module.exports = function(client, inbound, bridgeName) {
             else {
               var hangup = Q.denodeify(inbound.hangup.bind(inbound));
               hangup();
-              throw new CustomError('EarlyHangup', 
+              throw new CustomError('EarlyHangup',
                 dialed);
             }
           });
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -4,3 +4,5 @@ Co-op implementation of SLA in ARI using Node.js
 This implementation assumes that you have at least Asterisk 12.0.0 running. This is the version where ARI first came around, and it should have all of the functions required for this project.
 
 You must also have a valid ARI user in ari.conf named "user" and have a password "pass", as well as have 127.0.0.1 (or localhost) and 8088 configured as the bindaddr and bindport respectively in http.conf
+
+You must also have a dialplan extension in extensions.conf that leads to the application (must have same name as what application is being started in the code) and that has an argument to represent the SLA bridge to reach.
```
samuelg: I would include an example dialpan here. See http://spec.commonmark.org/0.17/#code-fence for how to include code in markdown.

app.js
```diff
@@ -20,10 +20,20 @@ connect('http://127.0.0.1:8088', 'user', 'pass')
...
-    .catch(errHandler)
-    .done();
+  var defer = Q.defer();
+  client.on('StasisStart', function(event, channel) {
+    stasisStart(event, client, channel);
+  });
+}
+
+function stasisStart (event, client, channel) {
+  if(event.args[0] !== 'dialed') {
```
samuelg: 1- That's not actually checking that the argument is numeric. That's just the example given in the standard ARI example. There's no guarantee you'll even see this in a real application. Look for a way to check that the argument is numeric.

2 - What happens if it's not? Right now your application would do nothing (that may be what you intend of course).

Nevermind on the first point, I see where you are controlling that to skip handling dialed channels.

app.js
```diff
@@ -20,10 +20,20 @@ connect('http://127.0.0.1:8088', 'user', 'pass')
...
-    .then(console.log)
-    .catch(errHandler)
-    .done();
+  var defer = Q.defer();
+  client.on('StasisStart', function(event, channel) {
+    stasisStart(event, client, channel);
+  });
+}
+
+function stasisStart (event, client, channel) {
```
samuelg: Instead of adding client in between the 2 StasisStart parameters, just move this function inside clientLoaded and reuse the client given there.

lib/sla.js
```diff
@@ -7,22 +7,25 @@ var Q = require('q');
...
-function designateMixingBridge(client) {
+function designateMixingBridge(client, bridgeNum) {
   var defer = Q.defer();
   var list = Q.denodeify(client.bridges.list.bind(client));
   list().then(
       function(bridges) {
         var bridge = bridges.filter(function(candidate) {
-          return candidate['bridge_type'] === 'mixing';
+          return (candidate['bridge_type'] === 'mixing' &&
+            candidate['name'] === bridgeNum);
```
samuelg: If this is really a bridge name, why name it bridgeNum?

lib/sla.js
```diff
@@ -84,12 +123,38 @@ function createChannel(client, bridge) {
...
-  return designateMixingBridge(client)
-    .then(function (bridge) {
-      return createChannel(client, bridge)
-        .then(function (channel) {
-          return addChannelsToBridge(channel, bridge);
-        });
-    });
+module.exports = function(client, channel, bridgeNumber) {
+  //This specifies what bridge number to use for SLA
+  if(!isNaN(parseInt(bridgeNumber))) {
```
samuelg: I would move this to a separate utility function.

lib/sla.js
```diff
@@ -84,12 +123,38 @@ function createChannel(client, bridge) {
...
+    var answer = Q.denodeify(channel.answer.bind(channel));
+    answer();
```
samuelg: This may or may not be done running before you run the next ARI request since it's not being used with a .then clause.

lib/sla.js
```diff
@@ -65,16 +91,29 @@ function addChannelsToBridge(channel, bridge) {
...
-  originate({endpoint: 'SIP/phone', app: 'sla'})
+  var dialed = client.Channel();
+  var originate = Q.denodeify(dialed.originate.bind(dialed));
+  originate({endpoint: 'SIP/phone', app: 'sla', appArgs: 'dialed'})
     .catch(function (err) {
       defer.reject(err);
     });
-  channel.once('StasisStart', function(event, channel) {
-    defer.resolve(channel);
+  dialed.on('StasisStart', function(event, dialed) {
```
samuelg: You should use once here. This will never happen more than once. Currently your event handler is going to stay in memory for the entire run of the program, essentially causing a memory leak.

lib/sla.js
```diff
@@ -65,16 +91,29 @@ function addChannelsToBridge(channel, bridge) {
...
+    var answer = Q.denodeify(dialed.answer.bind(dialed));
+    answer();
+    defer.resolve(dialed);
+  });
+  dialed.on('ChannelDestroyed', function(event, dialed) {
```
samuelg: Use once.

lib/sla.js
```diff
@@ -65,16 +91,29 @@ function addChannelsToBridge(channel, bridge) {
...
+    defer.resolve(null);
+  });
+  inbound.on('ChannelHangupRequest', function(event, inbound) {
```
samuelg: Use once.

lib/sla.js
```diff
@@ -65,16 +91,29 @@ function addChannelsToBridge(channel, bridge) {
...
+    defer.resolve(null);
+    var hangup = Q.denodeify(dialed.hangup.bind(dialed));
+    hangup()
+      .catch(function (err) {
+        defer.reject(err);
```
samuelg: You've already resolved the deferred before this, so this won't ever get called. You should resolve in the success callback of the .done clause so it only runs if catch wasn't called.

lib/sla.js
```diff
@@ -84,12 +123,38 @@ function createChannel(client, bridge) {
...
  */
-module.exports = function(client) {
-  return designateMixingBridge(client)
-    .then(function (bridge) {
-      return createChannel(client, bridge)
-        .then(function (channel) {
-          return addChannelsToBridge(channel, bridge);
-        });
-    });
+module.exports = function(client, channel, bridgeNumber) {
```
samuelg: You should rename channel to inbound to keep consistent with your naming scheme.

lib/sla.js
```diff
@@ -84,12 +123,38 @@ function createChannel(client, bridge) {
...
+              return addChannelsToBridge(channel, dialed, bridge);
+            }
+            else {
+              var hangup = Q.denodeify(channel.hangup.bind(channel));
+              hangup(function(channel) {
+                 console.log('Hanging up inbound channel %s', channel.id);
+              });
+              var defer = Q.defer();
+              defer.resolve('Channel rejected or caller hung up');
+              return defer.promise;
```
samuelg: You are already inside a promise chain, you can just return the string 'Channel rejected or caller hung up' if that is the result you want resolved.

lib/sla.js
```diff
@@ -84,12 +123,38 @@ function createChannel(client, bridge) {
...
+  }
+  else {
+    var defer = Q.defer();
+    var hangup = Q.denodeify(channel.hangup.bind(channel));
+    hangup()
+      .catch(function (err) {
+        defer.reject(err);
+      });
+    defer.resolve('Not a numeric SLA specification: ' +  bridgeNumber);
+    return defer.promise;
```
samuelg: If you just want to return a promise that will resolve to the string 'Not a numeric SLA specification: ' + bridgeNumber then just return Q('Not a numeric SLA specification: ' + bridgeNumber). No need for a deferred.

lib/sla.js
```diff
@@ -65,16 +91,29 @@ function addChannelsToBridge(channel, bridge) {
  *   addChannelsToBridge().
  * @return {Q} Q promise object.
  */
-function createChannel(client, bridge) {
```
samuelg, line 68: I don't think createChannel is a very good name for this function anymore since you are doing a lot more than just originating a channel. The fact that you are passing in an existing channel also makes the name very confusing.

tests/sla_tests.js
```diff
@@ -113,11 +116,27 @@ var getMockChannel = function() {
...
-          self.emit('StasisStart', {channel: {id: self.id}}, {channel: self});
-          pastStasis = true;
+          self.emit('StasisStart', {channel: {id: self.id}}, self);
+          pastOriginate = true;
         }
       }, asyncDelay);
     };
+    this.hangup = function(cb) {
+      var self = this;
+      for (var i = 0; i < channels.length; i++) {
```
samuelg: You can use channels.filter() to filter out the hungup channel. It would make the code a lot more readable.

tests/sla_tests.js
```diff
@@ -113,11 +116,27 @@ var getMockChannel = function() {
...
       setTimeout(function() {
         if(validEndpoint === input.endpoint) {
-          self.emit('StasisStart', {channel: {id: self.id}}, {channel: self});
-          pastStasis = true;
+          self.emit('StasisStart', {channel: {id: self.id}}, self);
+          pastOriginate = true;
         }
       }, asyncDelay);
     };
+    this.hangup = function(cb) {
```
samuelg, line 141: The callback never gets called.

tests/sla_tests.js
```diff
@@ -113,11 +116,27 @@ var getMockChannel = function() {
...
+        }
+      }
+        setTimeout(function() {
+          self.emit('ChannelHangupRequest', {channel: {id: self.id}},
+            {channel: self});
+          self.emit('ChannelDestroyed', {channel: {id: self.id}},
+            {channel: self});
+        }, (asyncDelay/10));
+    };
+    this.answer = function(cb) {
```
samuelg, line 155: The callback never gets called.

tests/sla_tests.js
```diff
@@ -206,4 +228,42 @@ describe('SLA Bridge and Channel Tester', function() {
...
+  it('should enter the application but specify an invalid SLA bridge',
+     function(done) {
+    var client = getMockClient();
+    var inbound = getMockChannel();
+    var sla = require('../lib/sla.js')(client, inbound, 'invalid').done();
+
+    incorrectBridge();
+    function incorrectBridge() {
+      setTimeout(function() {
+        if(!pastOriginate && bridgeChannels.length === 0 &&
```
samuelg: All those conditions are true after a call to afterEach so that's not actually testing what you are looking for. You have to have a way to flag that even though it was rejected, the code made it to a certain point.

tests/sla_tests.js
```diff
@@ -206,4 +228,42 @@ describe('SLA Bridge and Channel Tester', function() {
...
+     function(done) {
+    var client = getMockClient();
+    var inbound = getMockChannel();
+    var sla = require('../lib/sla.js')(client, inbound, '999').done();
+    channels[0].hangup();
+
+    earlyHangup();
+    function earlyHangup() {
+      setTimeout(function() {
+        if(!pastOriginate && bridgeChannels.length === 0 && isMixing &&
```
samuelg: You should verify that hangup() was called on inbound.

app.js
wolfeman67: 1. That is actually what I intend. If it has argument of 'dialed', then it should do nothing. Everything else gets a pass, (but it's checked if it's numeric later).

app.js
```diff
@@ -14,16 +13,21 @@ connect('http://127.0.0.1:8088', 'user', 'pass')
...
  *   connection.
  */
 function clientLoaded (client) {
   client.start('sla');
-  sla(client)
-    .then(console.log)
-    .catch(errHandler)
-    .done();
+  client.on('StasisStart', function(event, channel) {
+    if(event.args[0] !== 'dialed') {
```
samuelg: I would consider moving this check to a helper method. That way you can easily modify what determines whether or not a channel was dialed.

lib/sla.js
```diff
@@ -34,62 +38,133 @@ function designateMixingBridge(client) {
...
     .catch(function (err) {
       defer.reject(err);
+      
     });
-  channel.once('StasisStart', function(event, channel) {
-    defer.resolve(channel);
+  dialed.once('StasisStart', function(event, dialed) {
+    var answer = Q.denodeify(dialed.answer.bind(dialed));
+    answer();
+    defer.resolve(dialed);
```
samuelg, line 116: You can't resolve the promise on both StasisStart and ChannelDestroyed. Once a promise has been resolved it's locked. There's no way for the caller to call .then for ChannelDestroyed.

Update: you are still resolving the deferred twice for the dialed channel. At the very least you should have a comment to explain why you are doing this since it's not intuitive.

lib/sla.js
```diff
@@ -34,62 +38,133 @@ function designateMixingBridge(client) {
...
+      }).then(function (bridge) {
+        return originateChannel(client, inbound, bridge)
+          .then(function (dialed) {
+            if (dialed) {
+              return addChannelsToBridge(inbound, dialed, bridge);
+            }
+            else {
+              var hangup = Q.denodeify(inbound.hangup.bind(inbound));
+              hangup();
+              return 'Channel rejected or caller hung up';
```
samuelg: Returning a string for this is confusing. You should return a typed Error object. The caller can check the type to determine what to do with it. In other words, the program doesn't have to blow up, but returning a status as a string is error prone.

lib/sla.js
```diff
@@ -34,62 +38,133 @@ function designateMixingBridge(client) {
...
+            }
+          });
+      });
+  }
+  else {
+    var hangup = Q.denodeify(inbound.hangup.bind(inbound));
+    hangup();
+    return Q.reject(new Error('Not a numeric SLA specifcation: ' + bridgeName));
```
samuelg: s/specifcation/specification/

lib/sla.js
```diff
@@ -34,62 +38,133 @@ function designateMixingBridge(client) {
...
+  function channelLeft(event, object) {
+    numInBridge -= 1;
+    console.log('Channel %s has left the bridge', object.channel.id);
+    if(numInBridge === 1) {
+      if(object.channel.id === inbound.id) {
+        console.log('Hanging up dialed channel %s', dialed.id);
+        var hangupDialed = Q.denodeify(dialed.hangup.bind(dialed));
+        hangupDialed()
+          .catch(function (err) {
+            defer.reject(err);
```
samuelg, line 71: I would type this error so the caller can handle it without blowing up the program. It's good to know that a hangup failed but it should not blow up the entire program considering the channel has left at this point.

lib/sla.js
```diff
@@ -34,62 +38,133 @@ function designateMixingBridge(client) {
...
+          });
+      }
+      else if(object.channel.id === dialed.id) {
+        console.log('Hanging up inbound channel %s', inbound.id);
+        var hangupInbound = Q.denodeify(inbound.hangup.bind(inbound));
+        hangupInbound()
+          .catch(function (err) {
+            defer.reject(err);
```
samuelg, line 80: I would type this error so the caller can handle it without blowing up the program. It's good to know that a hangup failed but it should not blow up the entire program considering the channel has left at this point.

app.js
```diff
@@ -14,16 +13,21 @@ connect('http://127.0.0.1:8088', 'user', 'pass')
...
+      var bridgeName = event.args[0];
+      sla(client, channel, bridgeName)
+        .then(console.log)
+        .catch(errHandler)
```
samuelg, line 35: You should handle different errors differently. Not all errors thrown from sla should necessarily result in your program blowing up.

tests/sla_tests.js
```diff
@@ -110,13 +125,36 @@ var getMockChannel = function() {
...
-      }, asyncDelay);
+        else {
+          cb(new Error(self.id + ' is an invalid channel'));
+        }
+      }, answeringDelay);
+    };
+    this.hangup = function(cb) {
+      var self = this;
+      this['hungup'] = 'yes';
+      cb(null);
```
samuelg: You should call the callback in the setTimeout function.

tests/sla_tests.js
```diff
@@ -206,4 +252,51 @@ describe('SLA Bridge and Channel Tester', function() {
...
+      .catch(errHandler)
+      .done();
+    answeringDelay = 4 * asyncDelay;
+    inbound.hangup(function(){});
+
+    earlyHangup();
+    function earlyHangup() {
+      setTimeout(function() {
+        if(isMixing && channels.length === 0 && bridges.length === 1 &&
+          inbound['inbound'] && inbound['answered'] && inbound['hungup'] &&
```
samuelg: Since there's no _ in the property names, you can just use inbound.inbound, inbound.answered... Also, consider the names of boolean properties to clearly indicate their use. inbound.wasAnswered instead of inbound.answered.

samuelg:
Your tests are a lot easier to follow now that you store an array of channels with boolean properties to status checking.

samuelg:
Running jshint shows the following error:

lib/sla.js
     87 |  };
            ^ Unnecessary semicolon.

samuelg:
Currently 3 out of 8 tests fail.

samuelg:
Looks good.

PR #10, opened 2015-03-12
Config

wolfeman67:


app.js
```diff
@@ -43,7 +44,8 @@ function clientLoaded (client) {
  * @param {Object} err - error from application.
  */
 function errHandler(err) {
-  if(err.name === 'EarlyHangup' || err.name === 'HangupFailure') {
+  if(err.name === 'EarlyHangup' || err.name === 'HangupFailure' ||
```
samuelg: That is already a bit unwieldy and will only grow, I would move that check to a function. fatalError() perhaps?

app.js
```diff
@@ -29,8 +29,9 @@ function clientLoaded (client) {
   client.start('sla');
   client.on('StasisStart', function(event, channel) {
     if(!isDialed(event.args[0])) {
-      var bridgeName = event.args[0];
-      sla(client, channel, bridgeName)
+      var extension = event.args[0];
+      var confFile = process.argv[2];
```
samuelg: What happens if the conf file is not provided by the CLI?

res/config.json
```diff
@@ -0,0 +1,19 @@
+{
+  "sharedExtensions": [
+  {
+    "42": {
+      "trunks": [
+        {"42":
```
samuelg: This is hard to read. I would look at prettifying the json file so it's in a common format.

tests/testConfigs/multipleEndpoints.json
```diff
@@ -0,0 +1,19 @@
+{                                                                                
+  "sharedExtensions": [                                                          
+  {                                                                              
+    "42": {                                                             
+      "trunks": [                                                                
+        {"42":                                                                   
+          {                                                                      
+            "stations": [                                                        
+            {"endpoint": "SIP/phone1"
```
samuelg: Look at formatting here as well.

tests/testConfigs/noEndpoints.json
```diff
@@ -0,0 +1,16 @@
+{                                                                                
+  "sharedExtensions": [                                                          
+  {                                                                              
+    "42": {                                                             
+      "trunks": [                                                                
+        {"42":
```
samuelg: Look at formatting here as well.

tests/testConfigs/singleEndpoint.json
```diff
@@ -0,0 +1,17 @@
+{                                                                                
+  "sharedExtensions": [                                                          
+  {                                                                              
+    "42": {                                                             
+      "trunks": [                                                                
+        {"42":
```
samuelg: Look at formatting here as well.

lib/dal.js
```diff
@@ -0,0 +1,86 @@
...
+ * @param {string} confFile - the path and filename to the configuration file
+ * @param {string} name - the name of the sharedExtension to access
+ * @return {Q} - Q promise object
+ */
+
+var getSharedExtension = function(confFile, name) {
+  var defer = Q.defer();
+  var sharedExtension = {};
+  sharedExtension.pos = -1;
+  return getConfig(confFile).then(function(data) {
```
samuelg: You are putting an implementation detail directly in the API you expose. If you need to add database support later, you'll have a lot more work to separate this out from the functionality of returning a shared extension object. Also, what if you wanted to support both config files and a database?

You should have an internal only API for getting the data that can be modified to support various back-ends instead of having this inside the API you expose.

lib/dal.js
```diff
@@ -0,0 +1,86 @@
...
+/** Attempts to access configuration file data for the shared extension
+ * @param {string} confFile - the path and filename to the configuration file
+ * @param {string} name - the name of the sharedExtension to access
+ * @return {Q} - Q promise object
+ */
+
+var getSharedExtension = function(confFile, name) {
+  var defer = Q.defer();
+  var sharedExtension = {};
+  sharedExtension.pos = -1;
```
samuelg: You could also have a boolean called found to determine this since -1 is rather cryptic.

lib/dal.js
```diff
@@ -0,0 +1,86 @@
...
+  this.trunks.forEach(function(element, i) {
+    if(element[trunkName]) {
+      trunk.getAllStations = getAllStations;
+      trunk.trunkName = trunkName;
+      trunk.stations = element[trunk.name].stations;
+      trunk.pos = i;
+      defer.resolve(trunk);
+    }
+  });
+  if(!trunk.pos) {
```
samuelg: I would use a boolean here.

lib/sla.js
```diff
@@ -98,46 +99,102 @@ function addChannelsToBridge(inbound, dialed, bridge) {
...
+    toKill.forEach(function (unanswered) {
+      unanswered.removeListener('ChannelDestroyed', dialedChannelHungup);
+      var hangup = Q.denodeify(unanswered.hangup.bind(unanswered));
+      hangup();
+    });
+    defer.resolve(line);
+  }
+
+  function dialedChannelHungup(event) {
+    numToHangup -= 1;
```
samuelg: Why not remove the dialed channel from the dialed array and checking the length of the array instead? You could also have a dialed object with an array of channels that get updated when hungup and a areAllHungup() method.

lib/sla.js
```diff
@@ -98,46 +99,102 @@ function addChannelsToBridge(inbound, dialed, bridge) {
...
+  }
+
+  trunkData.getAllStations()
+    .then(function (stations) {
+      if (stations.length) {
+        stations.forEach(function() {
+          dialed.push(client.Channel());
+        });
+        numToHangup = dialed.length;
+        dialed.forEach(function (channel, i, dialed) {
```
samuelg: i is not a very descriptive variable name. At the very least use index.

lib/sla.js
```diff
@@ -98,46 +99,102 @@ function addChannelsToBridge(inbound, dialed, bridge) {
...
+          originate.push(Q.denodeify(channel.originate.bind(dialed[i])));
+          originate[i]({endpoint: stations[i].endpoint, app: 'sla',
```
samuelg: Why keep the bound originate methods around?

lib/sla.js
```diff
@@ -98,46 +99,102 @@ function addChannelsToBridge(inbound, dialed, bridge) {
...
+    });
     defer.resolve('Inbound channel hungup');
-  });
+  }
+
+  trunkData.getAllStations()
+    .then(function (stations) {
+      if (stations.length) {
+        stations.forEach(function() {
+          dialed.push(client.Channel());
```
samuelg: If you create an object with a station and a dialed channel, you'll be able to just iterate over that and not use an index anymore.

lib/sla.js
```diff
@@ -98,46 +99,102 @@ function addChannelsToBridge(inbound, dialed, bridge) {
...
+        });
+        numToHangup = dialed.length;
+        dialed.forEach(function (channel, i, dialed) {
+          originate.push(Q.denodeify(channel.originate.bind(dialed[i])));
+          originate[i]({endpoint: stations[i].endpoint, app: 'sla',
+            appArgs: 'dialed', timeout: 10})
```
samuelg: This should be indented 2 more spaces.

lib/sla.js
```diff
@@ -98,46 +99,102 @@ function addChannelsToBridge(inbound, dialed, bridge) {
...
-function checkIfNum (string) {
-  if(!isNaN(parseInt(string))) {
-    return true;
-  }
-  else {
-    return false;
-  }
+function getTrunkData (configurationFile, extension) {
+  var defer = Q.defer();
+  dal.getSharedExtension(configurationFile, extension)
```
samuelg: There's no need for a deferred here. Just return dal.getSharedExtension and within it either return a value on success or throw an error if not.

Deferred should only be created explicitly if you are creating a promise supported API or if you want the result of an event listener firing to resolve a promise. Since dal.getSharedExtension is already a promise supported API there's no reason to create a deferred yourself.

res/config.json
```diff
@@ -0,0 +1,22 @@
+{
+  "sharedExtensions": [
+    {
+      "42": {
+        "trunks": [
+          {
+            "42": {
+              "stations": [
```
matt-jordan, line 8: While it may be better to deal with this when you start manipulating outbound dial, stations should not be conceptually a child of trunks. Trunks are a type of endpoint that participate in a shared extension; they are a separate concept from the stations that they serve.

lib/dal.js
```diff
@@ -0,0 +1,76 @@
...
+
+/** Attempts to access configuration file data for the shared extension
+ * @param {string} confFile - the path and filename to the configuration file
+ * @param {string} name - the name of the sharedExtension to access
+ * @return {Q} - Q promise object
+ */
+
+var getSharedExtension = function(confFile, name) {
+  var defer = Q.defer();
+  var sharedExtension = {};
```
matt-jordan, line 14: From a stylistic perspective, it is often nice to have a line break between your variable declarations and the code that acts on them.

``` javascript
var defer = Q.defer();
var sharedExtension = {};

...
```

lib/dal.js
```diff
@@ -0,0 +1,76 @@
...
+  sharedExtension.found = false;
+  return fetchJSON(confFile).then(function(data) {
+    data.sharedExtensions.forEach(function(extension, i) {
+      if(extension[name]) {
```
matt-jordan: Space between if and conditional:

``` javascript
if (extension[name]) {
```

app.js
```diff
@@ -38,12 +42,25 @@ function clientLoaded (client) {
   });
 }
 
+/** Utility function for seeing if an error is fatal (crashes the program)
+ * @param {Object} err - the error in question.
+ * @return {boolean} - if the error is fatal or not
+ */
+function isFatal(err) {
+  if(err.name === 'EarlyHangup' || err.anme === 'HangupFailure' ||
```
matt-jordan: Space between if and conditional:

``` javascript
if (err.name ...
```

app.js
```diff
@@ -29,8 +33,8 @@ function clientLoaded (client) {
   client.start('sla');
   client.on('StasisStart', function(event, channel) {
     if(!isDialed(event.args[0])) {
-      var bridgeName = event.args[0];
-      sla(client, channel, bridgeName)
+      var extension = event.args[0];
+      sla(client, channel, confFile, extension)
```
matt-jordan: Your parameter ordering here is a bit odd:

1: The ARI client (okay)
2: The channel entering the shared extension (passed in from Stasis, okay)
3: The config file for the shared extension (but what shared extension?)
4: The actual shared extension (passed in from Stasis...)

From an ease of use perspective, you should order your parameters such that they flow logically. One ordering would be:

``` javascript
client, confFile, channel, extension
```

In other words: lead with the things that are unlikely to ever change (client, confFile) and follow with the more mutable pieces of data.

app.js
```diff
@@ -6,11 +6,15 @@ var sla = require('./lib/sla.js');
...
-  .catch(errHandler)
-  .done();
+var confFile;
+if (confFile = process.argv[2]) {
+  connect('http://127.0.0.1:8088', 'user', 'pass')
+    .then(clientLoaded)
+    .catch(errHandler)
+    .done();
+} else {
+  console.error('No configuration file specified');
```
matt-jordan: Is it fatal to not provide or find a configuration file? I would imagine this should actually bail out of the application.

lib/sla.js
```diff
@@ -98,65 +99,129 @@ function addChannelsToBridge(inbound, dialed, bridge) {
...
-        return unanswered.id !== line.id;
-      });
-      toKill.forEach(function (unanswered) {
-        var hangup = Q.denodeify(unanswered.hangup.bind(unanswered));
-        hangup();
-      });
-      defer.resolve(line);
+  var dialed = [];
+
+  function dialedEnteredStasis(event, line) {
```
matt-jordan, line 116: Document your function

lib/sla.js
```diff
@@ -98,65 +99,129 @@ function addChannelsToBridge(inbound, dialed, bridge) {
...
-    dialed[index].once('ChannelDestroyed', function(event) {
-      originate.pop(dialed[index]);
-      if(originate.length === 0) {
-        defer.resolve('All dialed channels hungup');
-        var hangup = Q.denodeify(inbound.hangup.bind(inbound));
-        hangup();
+    defer.resolve(line);
+  }
+
+  var allAreHungup = function(dialed) {
```
matt-jordan, line 137: Document this function

lib/sla.js
```diff
@@ -98,65 +99,129 @@ function addChannelsToBridge(inbound, dialed, bridge) {
...
+   if (dialed.length) {
+     return true;
+   } else {
+     return false;
+   }
+  };
+
+  var findInDialed = function(channel) {
```
matt-jordan, line 147: Document this function

lib/sla.js
```diff
@@ -98,65 +99,129 @@ function addChannelsToBridge(inbound, dialed, bridge) {
...
+    var position = -1;
+    dialed.forEach(function(object, index) {
+      if(channel === dialed.channel) {
+        position = index;
+      }
+    });
+    return position;
+  };
+
+  function dialedChannelHungup(event) {
```
matt-jordan, line 162: Document this function

lib/sla.js
```diff
@@ -98,65 +99,129 @@ function addChannelsToBridge(inbound, dialed, bridge) {
...
+            'dialed.  Element not found.'));
+    }
+        
+    if(allAreHungup(dialed)) {
+      inbound.removeListener('ChannelHangupRequest', inboundHungup);
+      defer.resolve('Dialed channels hungup');
+    }
+  }
+
+  function inboundHungup(event, line) {
```
matt-jordan, line 184: Document this function

lib/fetchJSON.js
```diff
@@ -0,0 +1,15 @@
+var fs = require('fs');
+var Q = require('q');
+
+/** Reads from the configuration file and returns data
```
matt-jordan: From a stylistic perspective, the description should go on the next line:

``` javascript
/**
 * Read from the configuration file and returns data
```

This finding applies to multiple comment blocks

res/config.json
wolfeman67, line 8: I have an idea of how to fix this, but I'll fix it in the next issue.

app.js
```diff
@@ -38,12 +42,25 @@ function clientLoaded (client) {
...
+      }
+}
+
 /**
  * Handles errors found in application.
  * @param {Object} err - error from application.
  */
 function errHandler(err) {
-  if(err.name === 'EarlyHangup' || err.name === 'HangupFailure') {
+  if (isFatal(err)) {
```
samuelg: It looks like you only log the error if it is fatal, but blow up if it is not fatal.

app.js
```diff
@@ -38,12 +42,25 @@ function clientLoaded (client) {
   });
 }
 
+/** Utility function for seeing if an error is fatal (crashes the program)
+ * @param {Object} err - the error in question.
+ * @return {boolean} - if the error is fatal or not
+ */
+function isFatal(err) {
+  if (err.name === 'EarlyHangup' || err.anme === 'HangupFailure' ||
```
samuelg: Since it evaluates to a boolean already, you can just use:

return err.name === 'EarlyHangup' || err.anme === 'HangupFailure' || ....;

lib/sla.js
```diff
@@ -98,75 +99,156 @@ function addChannelsToBridge(inbound, dialed, bridge) {
...
+    defer.resolve(line);
+  }
+  /**
+   * Utility function for cheking if all dialed channels have been hungup
+   * @param {Array} dialed - the array of channels; the length is checked
+   * @return {boolean} - whether or not the dialed array is empty
+   */
+
+  var allAreHungup = function(dialed) {
+   if (dialed.length) {
```
samuelg: You can just use:

return dialed.length;

to get the same result. Although I don't think it's doing what you think it's doing. Any integer above 0 will evaluate to true, so if the array has channels that have not been hungup yet, it will evaluate that all channels have been hungup. You should do !dialed.length instead so it only evaluates to true once the array has had all its elements removed.

lib/sla.js
```diff
@@ -98,75 +99,156 @@ function addChannelsToBridge(inbound, dialed, bridge) {
...
+      return Q.resolve(trunk);
+    })
+    .catch(function(invalid) {
+      // This will only be called if trunk and extension are not the same
+      // Shouldn't be right now
+      return Q.reject(new CustomError('InvalidTrunkSpecification', invalid));
+    });
+  })
+  .catch(function(invalid) {
+    if (invalid === 'Invalid specified extension: ' + extension) {
```
samuelg: You should check invalid.name for the Exception reason.

lib/sla.js
```diff
@@ -183,37 +265,37 @@ function CustomError(name, message) {
...
+              }
+              else {
+                throw new CustomError('EarlyHangup',
+                  dialed);
+              }
+            });
+        });
+    })
+  .catch(function (err) {
+    if(err.message !== 'Inbound channel hungup') {
```
samuelg: You should use a custom error and check err.name instead. Error messages can change, error names should not.

samuelg:
Looks good. Make sure to squash into a single commit before merging.

PR #13, opened 2015-03-30
Allows for compatibility with BLF

TACambron:
Added updates throughout the code to update each phone's status in real time so that BLF can be properly used on the phones.

### DIFF
--- a/lib/sla.js
+++ b/lib/sla.js
@@ -46,7 +46,7 @@ function designateMixingBridge(client, bridgeName) {
  * @param {Object} bridge - Bridge that channels are to be added to.
  * @return {Q} Q promise object.
  */
-function addChannelsToBridge(inbound, dialed, bridge) {
+function addChannelsToBridge(client, inbound, dialed, bridge) {
   var defer = Q.defer();
   var numInBridge = 0;
   console.log('Adding channels to bridge %s', bridge.id);
@@ -87,6 +87,9 @@ function addChannelsToBridge(inbound, dialed, bridge) {
       defer.resolve('Application completed');
       object.bridge.removeListener('ChannelEnteredBridge', channelEntered);
       object.bridge.removeListener('ChannelLeftBridge', channelLeft);
+      var deviceState = Q.denodeify(client.deviceStates.update.bind(client));
+      deviceState({deviceName: 'Stasis:' + object.bridge.name,
+        deviceState: 'NOT_INUSE'});
     }
   }
   bridge.on('ChannelLeftBridge', channelLeft);
@@ -116,6 +119,8 @@ function originateChannel(client, inbound, bridge, trunkData) {
   function dialedEnteredStasis(event, line) {
     var answer = Q.denodeify(line.answer.bind(line));
     answer();
+    var deviceState = Q.denodeify(client.deviceStates.update.bind(client));
+    deviceState({deviceName: 'Stasis:' + bridge.name, deviceState: 'INUSE'});
     var toKill = dialed.filter(function (unanswered) {
       return unanswered.channel.id !== line.id;
     });
@@ -125,6 +130,9 @@ function originateChannel(client, inbound, bridge, trunkData) {
       var hangup = Q.denodeify(unanswered.channel.hangup.bind(
           unanswered.channel));
       hangup();
+      var deviceState = Q.denodeify(client.deviceStates.update.bind(client));
+      deviceState({deviceName: 'Stasis:' + bridge.name,
+        deviceState: 'NOT_INUSE'});
     });
     defer.resolve(line);
   }
@@ -170,6 +178,9 @@ function originateChannel(client, inbound, bridge, trunkData) {
         
     if (allAreHungup(dialed)) {
       inbound.removeListener('ChannelHangupRequest', inboundHungup);
+      var deviceState = Q.denodeify(client.deviceStates.update.bind(client));
+      deviceState({deviceName: 'Stasis:' + bridge.name,
+        deviceState: 'NOT_INUSE'});
       defer.reject(new customError.CustomError('DialedHungup', 'Dialed ' +
             'channels hungup'));
     }
@@ -188,6 +199,9 @@ function originateChannel(client, inbound, bridge, trunkData) {
       var hangup = Q.denodeify(unanswered.channel.hangup.bind(
           unanswered.channel));
       hangup();
+      var deviceState = Q.denodeify(client.deviceStates.update.bind(client));
+      deviceState({deviceName: 'Stasis:' + bridge.name,
+        deviceState: 'NOT_INUSE'});
     });
     defer.reject(new customError.CustomError('InboundHungup',
           'Inbound channel hungup'));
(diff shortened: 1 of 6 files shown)


lib/sla.js
```diff
@@ -116,6 +116,8 @@ function originateChannel(client, inbound, bridge, trunkData) {
   function dialedEnteredStasis(event, line) {
     var answer = Q.denodeify(line.answer.bind(line));
     answer();
+    var deviceState = Q.denodeify(client.deviceStates.update.bind(client));
```
matt-jordan: From a readability perspective, put some whitespace between logical groupings of code:

```
var answer = ...
answer()

var deviceState = ...
deviceState(...)

var toKill = ...
```

etc.

matt-jordan: Note that this comment applies throughout this review. You've got a lot of code bunched up, without much in the way marking logical separation between actions being taken.

lib/sla.js
```diff
@@ -125,6 +127,9 @@ function originateChannel(client, inbound, bridge, trunkData) {
       var hangup = Q.denodeify(unanswered.channel.hangup.bind(
           unanswered.channel));
       hangup();
+      var deviceState = Q.denodeify(client.deviceStates.update.bind(client));
```
matt-jordan: This pattern of getting the device state from the client, setting up the device to change, and changing its state, is pretty common throughout this patch.

I'd write a small function that does just that, taking in the name of the device to change and the state to set it to.

tests/sla_tests.js
```diff
@@ -429,4 +465,180 @@ describe('SLA Bridge and Channels Tester', function() {
       }, asyncDelay);
     } 
   });
+  it('should mark a device as RINGING when dialing an outbound channel',
```
matt-jordan, line 461: Vertical whitespace is your friend!

tests/sla_tests.js
```diff
@@ -429,4 +465,180 @@ describe('SLA Bridge and Channels Tester', function() {
...
+      function(done) {
+    var client = getMockClient();
+    var inbound = getMockChannel();
+    inbound.inbound = true;
```
matt-jordan: From a naming perspective, `inbound.inbound` is pretty weird. I'd name the channel something other than `inbound` to avoid the duplication.

tests/sla_tests.js
```diff
@@ -82,10 +84,42 @@ var getMockClient = function() {
         cb(null, bridges[bridges.length-1]);
       }
     };
-    this.Channel= function() {
+    this.Channel = function() {
       var newChan = getMockChannel();
       return newChan;
     };
+    this.deviceStates = {
+      update: function(param, cb) {
```
samuelg: 'param' should really be 'params' as this is an object. Another common name for such objects in JavaScript is 'options'.

tests/sla_tests.js
```diff
@@ -82,10 +84,42 @@ var getMockClient = function() {
...
+        var exists = mockDeviceStates.filter(function(deviceState) {
```
samuelg: 'exists' implies this is a boolean. I would rename it to something else since it's used later in to establish whether or not the array is empty.

Another options is to do:

var exists = !mockDeviceStates.filter(function(deviceState) {
  return deviceState.deviceName === param.deviceName;
}).length;

but that is definitely harder to read.

The best option is to use a library like https://lodash.com to overcome some of JavaScript limitations. You could can do something like:

var exists = _.some(mockDeviceStates, function(deviceState) {
  return deviceState.deviceName === param.deviceName;
});

tests/sla_tests.js
```diff
@@ -231,7 +267,7 @@ describe('SLA Bridge and Channels Tester', function() {
     var sla = require('../lib/sla.js')(client, config, inbound, '42')
       .done();
 
-    bridges.push(getMockBridge({type: 'mixing', name: '42'}, function(){}));
+    bridges.push(getMockBridge({type: 'mixing', name: '42'}, function() {}));
```
samuelg: I'm not sure why there's a function being passed to getMockBridge. That function doesn't expect more than 1 parameter.

tests/sla_tests.js
```diff
@@ -429,4 +465,180 @@ describe('SLA Bridge and Channels Tester', function() {
...
+          }
+      }, asyncDelay);
+    } 
+  });
+  it('should not mark a device INUSE when a trunk enters the shared extension',
+      function(done) {
+    var client = getMockClient();
+    var inbound = getMockChannel();
+    inbound.inbound = true;
+    bridges.push(getMockBridge({type: 'mixing', name: '42'}, function() {}));
```
samuelg: Same here as far as passing a function to getMockBridge.

tests/sla_tests.js
```diff
@@ -429,4 +465,180 @@ describe('SLA Bridge and Channels Tester', function() {
...
+    inbound.inbound = true;
+    config = 'tests/testConfigs/noEndpoints.json';
+    var sla = require('../lib/sla.js')(client, config, inbound, '42')
+      .catch(errHandler)
+      .done();
+
+    noEndpoints();
+    function noEndpoints() {
+      setTimeout(function() {
+        if (noEndpoints) {
```
samuelg: noEndpoints doesn't appear to be defined so it should always be undefined which will be evaluated to false.

tests/sla_tests.js
```diff
@@ -429,4 +465,180 @@ describe('SLA Bridge and Channels Tester', function() {
...
+    inbound.inbound = true;
+    config = 'tests/testConfigs/invalid.json';
+    var sla = require('../lib/sla.js')(client, config, inbound, '42')
+      .catch(errHandler)
+      .done();
+
+    invalidConfigurationFile();
+    function invalidConfigurationFile() {
+      setTimeout(function() {
+        if (configurationFailed) {
```
samuelg: Typically, using setTimeout to check local variables for state during a test is done because you have no direct control over the state you want to check against.

It would be better here to simply provide your own anonymous function to the catch handler instead of relying on errHandler. You could then call done() inside your anonymous function if the error thrown is the one you expect.

tests/sla_tests.js
```diff
@@ -429,4 +465,180 @@ describe('SLA Bridge and Channels Tester', function() {
...
+  });
+  it('should mark a device as INUSE when dialing an outbound channel',
+      function(done) {
+    var client = getMockClient();
+    var inbound = getMockChannel();
+    inbound.inbound = true;
+    var sla = require('../lib/sla.js')(client, config, inbound, '42')
+      .catch(errHandler)
+      .done();
+    client.deviceStates.update({deviceName: 'Stasis:42',
```
samuelg: This is hard to read, you should go for one of the approached below:

client.deviceStates.update({
  deviceName: 'Stasis:42',
  deviceState: 'NOT_INUSE'
}, function() {
});

client.deviceStates.update({deviceName: 'Stasis:42',
                                            deviceState: 'NOT_INUSE'}, function() {});

The important thing is to align subsequent lines in a way that flows better and is easier to read.

Note: github didn't indent my comment as it should have. :(

tests/sla_tests.js
```diff
@@ -429,4 +465,180 @@ describe('SLA Bridge and Channels Tester', function() {
...
+      deviceState: 'NOT_INUSE'}, function() {});
+    answeringDelay = 2 * asyncDelay;
+
+    markAsRinging();
+    function markAsRinging() {
+      setTimeout(function() {
+        if (isMixing && bridges.length === 1 && inbound.inbound &&
+          dialed[0] && dialed[0].wasAnswered && inbound.wasAnswered &&
```
samuelg: dialed[0] should be aligned with isMixing and so forth so this statement is easier to read.

samuelg:
+1 on everything Matt said. Vertical spacing would help a lot in making your code more readable.

samuelg:
Looks good. Make sure you squash the commits before merging.

TACambron:
Squashed into #18

PR #16, opened 2015-04-03
First real rough draft of outbound dialing

wolfeman67:
Allows a station to outbound dial, and also allows stations to
join a call if there is one in progress.  Kicks out additional
outside channels if one attempts to call the extension when a
call is in progress.  Checks device states to see if entry is
allowed.


lib/dal.js
```diff
@@ -12,26 +29,32 @@ var customError = require('./customError.js');
 var getSharedExtension = function(confFile, name) {
   var defer = Q.defer();
   var sharedExtension = {};
-
   sharedExtension.found = false;
-  return fetchJSON(confFile).then(function(data) {
-    data.sharedExtensions.forEach(function(extension, i) {
+  fetchJSON(confFile).then(function(data) {
```
samuelg: There's no reason to have a deferred here. fetchJSON already is async. Just return the result of that. Everywhere where you reject, just throw instead, and everywhere where you resolve return instead.

lib/dal.js
```diff
@@ -12,26 +29,32 @@ var customError = require('./customError.js');
...
+    data.sharedExtensions.some(function(extension, i) {
```
samuelg: I'm not sure where you are getting some(). It's not a built in JavaScript function and it looks like you are calling it on a built in array. It's part of lodash for example but you're not using it the way it's intended there.

lib/sla.js
```diff
@@ -40,54 +41,108 @@ function designateMixingBridge(client, bridgeName) {
...
+ * @param {Object} client - Object that contains information from the ARI
+ *   connection.  Houses the device states
+ * @param {String} name - name of the shared extension to access.
+ * @return {String} - the current device state of this extension
+ */
+function getState(client, name) {
+  var getDeviceState = Q.denodeify(client.deviceStates.get.bind(client
+        .deviceStates));
+  return getDeviceState({deviceName: 'Stasis:' + name}).then(function (ds) {
+    return Q.resolve(ds.state);
```
samuelg: You can simply return ds.state here.

lib/sla.js
```diff
@@ -40,54 +41,108 @@ function designateMixingBridge(client, bridgeName) {
...
- * @param {Object} inbound - Incoming channel to be added to the bridge.
+ * Updates the state of the shared extension.
+ * @param {Object} client - Object that contains information from the ARI
+ *   connection.
+ * @param {String} name - name of the shared extension to update.
+ * @param {String} state - state to update the shared extension to.
+ */
+function updateState(client, name, state) {
+  var deviceState = Q.denodeify(client.deviceStates.update.bind(client));
+  deviceState({deviceName: 'Stasis:' + name, deviceState: state});
```
samuelg: Only problem here is that since you don't return deviceState() there's no way for the caller to determine if an error occurred.

lib/sla.js
```diff
@@ -40,54 +41,108 @@ function designateMixingBridge(client, bridgeName) {
...
+      } else {
+        if (object.bridge.channels.length) {
+          console.log('Hanging up station channel/s');
+          object.bridge.channels.forEach( function(ID) {
+            var hangup = Q.denodeify(client.channels.hangup.bind
+              (client.channels));
+            hangup({channelId: ID})
+              .catch(function (err) {
+                err.name = 'HangupFailure';
+                defer.reject(err);
```
samuelg: This may or may not be a problem but rejecting on error here means you can't differentiate between a single channel erroring on hangup vs multiple.

lib/sla.js
```diff
@@ -40,54 +41,108 @@ function designateMixingBridge(client, bridgeName) {
...
+          hangupOutsider({channelId: object.bridge.channels[0]})
+            .catch(function (err) {
+              err.name = 'HangupFailure';
+              defer.reject(err);
+            });
+        }
+      } else {
+        if (object.bridge.channels.length) {
+          console.log('Hanging up station channel/s');
+          object.bridge.channels.forEach( function(ID) {
```
samuelg, line 142: This will work as expected but you may want to look at Q.all() to run the hangups concurrently. It will then call then once all have been completed or catch if a single async operation failed.

lib/sla.js
```diff
@@ -40,54 +41,108 @@ function designateMixingBridge(client, bridgeName) {
...
-      else if (object.channel.id === dialed.id) {
-        console.log('Hanging up inbound channel %s', inbound.id);
-        var hangupInbound = Q.denodeify(inbound.hangup.bind(inbound));
-        hangupInbound()
-          .catch(function (err) {
-            err.name = 'HangupFailure';
-            defer.reject(err);
-          });
+      if (object.bridge.channels.length === 0) {
+        defer.resolve(true);
```
samuelg: Why resolve the boolean value true? Resolve already signals success.

lib/sla.js
```diff
@@ -126,6 +179,7 @@ function originateChannel(client, inbound, bridge, trunkData) {
           unanswered.channel));
       hangup();
     });
+    updateState(client, bridge.name, 'INUSE');
```
samuelg: This should be changed as per Tyler's BLF pull request when rebasing if not already part of his pull request.

lib/sla.js
```diff
@@ -167,11 +221,11 @@ function originateChannel(client, inbound, bridge, trunkData) {
       defer.reject(new customError.CustomError('BadRemoval','Failed to remove' +
            ' element in dialed.  Element not found.'));
     }
-        
     if (allAreHungup(dialed)) {
+      updateState(client, bridge.name, 'NOT_INUSE');
```
samuelg: Same here. Hard coding state strings was fixed in Tyler's PR.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
           unanswered.channel));
       hangup();
     });
+    updateState(client, bridge.name, 'NOT_INUSE');
```
samuelg: Same here. Hard coding state strings was fixed in Tyler's PR.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
-          dialed.push({endpoint: station.endpoint, channel: client.Channel()});
-        });
-        dialed.forEach(function (object) {
-          var originate = Q.denodeify(object.channel.originate.bind(
-              object.channel));
-          originate({endpoint: object.endpoint, app: 'sla',
-              appArgs: 'dialed', timeout: 10})
-            .catch(function (err) {
-              defer.reject(err);
+  var defer = Q.defer();
```
samuelg, line 322: You could use vertical spacing throughout this pull request to make the code more readable. This section is especially dense.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+  var dialed = [];
+  getState(client, bridge.name).then(function(currentState) {
+    if (currentState === 'UNKNOWN') {
+      updateState(client, bridge.name, 'NOT_INUSE');
```
samuelg: Same here. Hard coding state strings was fixed in Tyler's PR.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+            });
+            dialed.forEach(function (object) {
+              var originate = Q.denodeify(object.channel.originate.bind(
+                  object.channel));
+              originate({endpoint: object.endpoint, app: 'sla',
+                  appArgs: 'dialed', timeout: 10})
+                .catch(function (err) {
+                  defer.reject(err);
+                });
+              updateState(client, bridge.name, 'RINGING');
```
samuelg: Same here. Hard coding state strings was fixed in Tyler's PR.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
-            .catch(function (err) {
-              defer.reject(err);
+  var defer = Q.defer();
+  var dialed = [];
+  getState(client, bridge.name).then(function(currentState) {
+    if (currentState === 'UNKNOWN') {
+      updateState(client, bridge.name, 'NOT_INUSE');
+      currentState = 'NOT_INUSE';
+    }
+    if (currentState === 'NOT_INUSE') {
```
samuelg: Same here. Hard coding state strings was fixed in Tyler's PR.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
-          originate({endpoint: object.endpoint, app: 'sla',
-              appArgs: 'dialed', timeout: 10})
-            .catch(function (err) {
-              defer.reject(err);
+  var defer = Q.defer();
+  var dialed = [];
+  getState(client, bridge.name).then(function(currentState) {
+    if (currentState === 'UNKNOWN') {
+      updateState(client, bridge.name, 'NOT_INUSE');
+      currentState = 'NOT_INUSE';
```
samuelg: Same here. Hard coding state strings was fixed in Tyler's PR.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
-        dialed.forEach(function (object) {
-          var originate = Q.denodeify(object.channel.originate.bind(
-              object.channel));
-          originate({endpoint: object.endpoint, app: 'sla',
-              appArgs: 'dialed', timeout: 10})
-            .catch(function (err) {
-              defer.reject(err);
+  var defer = Q.defer();
+  var dialed = [];
+  getState(client, bridge.name).then(function(currentState) {
```
samuelg: getState is already async, there's no reason to use a deferred here. You can simply return when you have something to return or on success and throw when there's an error. Deferred should only be used when implementing an async method that doesn't already rely on an async function.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+      var kickOutUser = Q.denodeify(inbound.continueInDialplan.bind(inbound));
+      kickOutUser();
+      defer.reject(new customError.CustomError('ExtensionOccupied',
+            'An inbound caller attempted to call into a busy extension'));
+    }
+  });
+  return defer.promise;
+}
+
+function originateOutbound(client, outbound, bridge, extension, trunkData) {
```
samuelg: I would reorder the parameters here from most general to specific. client, extension, trunkData, outbound, bridge. Keep things that change often at the end of the function signature.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+  /**
+   * The function that gets called once the dialed channel enters Stasis.
+   *   Makes the extension busy and clears eventListeners
+   * @param {Object} event - the information related to the StasisStart event
+   * @param {Object} line - the line that has entered Stasis
+   */
+  function dialedEnteredStasis(event, line) {
+    updateState(client, bridge.name, 'INUSE');
```
samuelg: Same here. Hard coding state strings was fixed in Tyler's PR.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+    answer();
+    defer.resolve(line);
+  }
+  
+  /**
+   * Function that gets called when a dialed channel hangs up
+   * @param {Object} event - the event object related to this hang up
+   */
+  function dialedChannelHungup(event) {
+    updateState(client, bridge.name, 'NOT_INUSE');
```
samuelg: Same here. Hard coding state strings was fixed in Tyler's PR.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+      'Dialed channel hungup'));
+  }
+  /**
+   * Function that gets called when the outbound caller hangs up with a
+   * calling in progress
+   * @param {Object} event - the event object related to this hang up
+   * @param {line} line - the channel that hungup
+   */
+  function outboundHungup(event, line) {
+    updateState(client, bridge.name, 'NOT_INUSE');
```
samuelg: Same here. Hard coding state strings was fixed in Tyler's PR.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+          .catch(function(err) {
+            defer.reject(err);
+          });
+        updateState(client, bridge.name, 'RINGING');
+        channelToDial.once('StasisStart', dialedEnteredStasis);
+        channelToDial.once('ChannelDestroyed', dialedChannelHungup);
+        outbound.removeListener('ChannelHangupRequest', outboundHungupEarly);
+        outbound.once('ChannelHangupRequest', outboundHungup);
+        break;
+    default:
```
samuelg: Typically better to align the default case with previous case statements.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+        var originate = Q.denodeify(channelToDial.originate.bind(
+              channelToDial));
+        originate({endpoint: 'SIP/' + toDial + '@'  +
+          trunkData[0], app: 'sla',
+          appArgs: 'dialed',
+          timeout: 10})
+          .catch(function(err) {
+            defer.reject(err);
+          });
+        updateState(client, bridge.name, 'RINGING');
```
samuelg: Same here. Hard coding state strings was fixed in Tyler's PR.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+   */
+  function dtmfReceived(event, channel) {
+    var digit = event.digit;
+    switch (digit) {
+      case '#':
+        channelToDial = client.Channel();
+        var originate = Q.denodeify(channelToDial.originate.bind(
+              channelToDial));
+        originate({endpoint: 'SIP/' + toDial + '@'  +
+          trunkData[0], app: 'sla',
```
samuelg: trunkData should be aligned with endpoint.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+      toDial += digit;
+      break;
+    }
+  }
+  var toDial = '';
+  var callInProgress = false;
+  var channelToDial;
+  var defer = Q.defer();
+  getState(client, bridge.name).then(function (currentState) {
+    if (currentState === 'INUSE' || currentState === 'RINGING' ||
```
samuelg: Same here. Hard coding state strings was fixed in Tyler's PR.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+      currentState === 'BUSY') {
+      callInProgress = true;
+    }
+    if (!trunkData[0]) {
+      defer.reject(new customError.CustomError('NoTrunks', 'There are no ' +
+         'trunks with which to outbound dial'));
+    } else if (!callInProgress) {
+      var playback = client.Playback();
+      outbound.play({media: 'sound:pls-entr-num-uwish2-call'}, playback);
+      updateState(client, bridge.name, 'BUSY');
```
samuelg: Same here. Hard coding state strings was fixed in Tyler's PR.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+    default:
+      toDial += digit;
+      break;
+    }
+  }
+  var toDial = '';
+  var callInProgress = false;
+  var channelToDial;
+  var defer = Q.defer();
+  getState(client, bridge.name).then(function (currentState) {
```
samuelg: getState is already async, there's no reason to use a deferred here.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
-    })
-    .catch(function(invalid) {
-      // This will only be called if trunk and extension are not the same
-      // Shouldn't be right now
-      return Q.reject(new customError.CustomError('InvalidTrunkSpecification',
-          invalid));
+function isStation (extension, channel) {
+  return extension.getAllStations().then(function(stations) {
+    var isStation = false;
+    stations.some(function(station) {
```
samuelg: Again, not sure why you are calling some() here.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+      var re = regexp().find(station).toRegExp();
+      // First conditional is for regular SIP users like SIP/phone1
+      if (re.test(channel.name)) {
+        isStation = true;
+        return true;
```
samuelg: Why return the boolean value true? Calling Return; will signal success without having to pass that value. Error conditions should be signaled with throw instead.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
         });
+      });
     })
   .catch(function (err) {
-    if(err.name !== 'InboundHungup') {
-      var hangup = Q.denodeify(inbound.hangup.bind(inbound));
-      hangup();
-    }
+    if (err.name !== 'InboundHungup' && err.name !== 'OutboundHungup' &&
+      err.name !== 'OutboundHungupEarly' && err.name !== 'ExtensionOccupied') {
```
samuelg: err.name should be aligned with the one above it.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+              return addChannelsToBridge(client, channel, dialed,
+                bridge, extension).then(function(toFree) {
+                  if (toFree === true) {
+                    return Q.resolve('Extension ' + extensionName + ' freed!');
+                  }
+                });
             });
+            } else {
+              return originateOutbound(client, channel, bridge, extension,
+                trunks)
```
samuelg: trunks should be aligned with client.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+                      if (toFree === true) {
+                        return Q.resolve('Extension ' + extensionName +
+                          ' freed!');
+                      }
+                    });
+                })
+              .catch(function (err) {
+                if (err.name === 'ExtensionBusy') {
+                  return addChannelsToBridge(client, channel, null, bridge,
+                    extension).then(function(toFree) {
```
samuelg: extension should be aligned with client.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+          return answer()
+          .then(function() {
+            return designateMixingBridge(client, extensionName);
+          }).then(function (bridge) {
+            if (!isStation) {
+              return originateInbound(client, channel, bridge, extension)
             .then(function (dialed) {
-              return addChannelsToBridge(inbound, dialed, bridge);
+              return addChannelsToBridge(client, channel, dialed,
+                bridge, extension).then(function(toFree) {
```
samuelg: bridge should be aligned with client.

lib/sla.js
```diff
@@ -189,96 +243,276 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
-    .then(function(trunk) {
-      var answer = Q.denodeify(inbound.answer.bind(inbound));
-      return answer()
-        .then(function() {
-          return designateMixingBridge(client, extension);
-        }).then(function (bridge) {
-          return originateChannel(client, inbound, bridge, trunk)
+module.exports = function(client, confFile, channel, extensionName) {
+  return dal.getSharedExtension(confFile, extensionName)
+    .then(function(extension) {
```
samuelg: This is getting really unwieldy. I would do some research on ways to limit nested promises.

tests/sla_tests.js
```diff
@@ -385,46 +493,505 @@ describe('SLA Bridge and Channels Tester', function() {
...
+              hangupRequested = true;
+            channel.hangup(function (){});
+            secondChannel.hangup(function (){});
+          }
+          bothStationsHangup();
+        }
+      }, asyncDelay);
+    } 
+  });
+  it('should test whether the application keeps the non-station channel with ' +
```
samuelg, line 911: There should be a blank line between tests here. Overall your pull request could really do with more vertical spacing to help readability.

update: There should be a space here between tests.

samuelg:
Looks like the old commits from before you rebased off of master are back in. Not sure how that happened but this branch should be cleaned up so it does not contain them anymore.

tests/sla_tests.js
```diff
@@ -385,46 +508,513 @@ describe('SLA Bridge and Channels Tester', function() {
...
+            !hangupRequested) {
+              hangupRequested = true;
+            channel.hangup(function (){});
+          }
+          oneStationHangsup();
+        }
+      }, asyncDelay);
+    } 
+  });
+  it('should hangup both station channels when the outside channel hangs up',
```
samuelg, line 968: There should be a space here between tests.

PR #19, opened 2015-04-08
Outbound single

wolfeman67:

PR #20, opened 2015-04-08
Hold basic

wolfeman67:
Now adds the ability for this application to intercept hold signals from the stations (they have to be in the same Asterisk instance).  If all stations in the shared extension have either pressed the hold button or hung up, music on hold will be played on the outside user's device.  If a station dials in again while music on hold is being played, then the outside user will be put off hold, music on hold will cease, and conversations between station/s and an outside user continue as normal.  The only for an extension to be freed is if the outside user hangs up.


samuelg:
The pull request has no description. What is the intent for these additions? What is the expected behavior?

README.md
```diff
@@ -5,37 +5,54 @@ This implementation assumes that you have at least Asterisk 12.0.0 running. This
...
+          "SIP/phone1",
+          "SIP/phone2"                                            
+        ]                                                                
+      }                                                                            
+    }                                                                              
   ]                                                                              
 }
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-Right now, however, this application only supports one trunk per sharedExtension, and the sharedExtension MUST have the same name as the trunk. The support for multiple trunks (such as 42A and 42B) will come later.
+Right now, however, this application only supports one trunk per sharedExtension, it can have any name, however, but the SIP trunk must appear in sip.conf, like below:
```
matt-jordan, line 27: Run-on sentence. I'd also avoid colloquialisms, such as "right now."

matt-jordan, line 27: Break this section out into an example sip.conf configuration.

README.md
```diff
@@ -5,37 +5,54 @@ This implementation assumes that you have at least Asterisk 12.0.0 running. This
...
+type = peer
+host = 0.0.0.0
+username = 42-A
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-You must also have a dialplan extension in extensions.conf that leads to the application (must have same name as what application is being started in the code) and that has an argument to represent the SLA bridge to reach.  An example is below:
+You must also have a dialplan extension in extensions.conf that leads to the application (must have same name as what application is being started in the code) and that has an argument to represent the SLA bridge to reach.  An example with a device state hint is below:
 
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+exten => 42,hint,Stasis:42
```
matt-jordan, line 38: Break this section out into a sample extensions.conf configuration.

README.md
```diff
@@ -5,37 +5,54 @@ This implementation assumes that you have at least Asterisk 12.0.0 running. This
...
+Right now, however, this application only supports one trunk per sharedExtension, it can have any name, however, but the SIP trunk must appear in sip.conf, like below: 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+[42-A]
+type = peer
+host = 0.0.0.0
+username = 42-A
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-You must also have a dialplan extension in extensions.conf that leads to the application (must have same name as what application is being started in the code) and that has an argument to represent the SLA bridge to reach.  An example is below:
+You must also have a dialplan extension in extensions.conf that leads to the application (must have same name as what application is being started in the code) and that has an argument to represent the SLA bridge to reach.  An example with a device state hint is below:
```
matt-jordan, line 35: Get in the habit of only using one space between sentences. I'd also remove the paranthetical aside, as it isn't a paranthetical. It is necessary to have the same name as the application name in the node.js app.

README.md
```diff
@@ -5,37 +5,54 @@ This implementation assumes that you have at least Asterisk 12.0.0 running. This
...
+exten => 42,1,NoOp()                                                             
+    same => n,Stasis(sla,42)
+    same => n,Goto(420,1)                                                 
+
+exten => 420,hint,Stasis:420
+exten => 420,1,NoOp()
+    same => n,Stasis(sla,420)
+    same => n,Hangup()
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The ARI patch for detecting channel hold events is also required (may be in the Asterisk 13 branch by the time we are finished).
```
matt-jordan, line 56: Remove this sentence.

Instead, note that the minimum version of Asterisk required for this ARI application is 13.4.0.

app.js
```diff
@@ -4,11 +4,12 @@ var ari = require('ari-client');
 var util = require('util');
 var sla = require('./lib/sla.js');
 var Q = require('q');
+var dal = require('./lib/dal.js');
 
 var connect = Q.denodeify(ari.connect);
 var confFile;
 if (confFile = process.argv[2]) {
-  connect('http://127.0.0.1:8088', 'user', 'pass')
+  connect('http://localhost:8088', 'user', 'pass')
```
matt-jordan, line 12: I'd revert this change, and/or do it in a separate pull request.

Your pull requests should focus on solving the problem at hand: in this case, implementing call hold. Ancillary changes, such as switching from a specific IPv4 representation of a local loopback to a hostname resolution for a local machine is ancillary to your task, and should be done as a separate commit/pull request.

app.js
```diff
@@ -34,6 +35,7 @@ function clientLoaded (client) {
   client.on('StasisStart', function(event, channel) {
     if (!isDialed(event.args[0])) {
       var extension = event.args[0];
+      var confFile = process.argv[2];
```
matt-jordan, line 38: Why would this not be extracted prior to StasisStart? As it is, it looks like you are extracting the confFile for every single channel that enters into the application. You should only be doing it once, when your client loads.

lib/dal.js
```diff
@@ -2,35 +2,60 @@ var Q = require('q');
...
+  this.name = name;
+  this.message = message;
+}
+
 /** 
  * Attempts to access configuration file data for the shared extension
+ *   If this is the first time accessing this said extension, then it creates
+ *   a new sharedExtension "object" and its relevant underlying data structures,
+ *   adds it to the array of extsInUse and returns it.
+ *   Simply returns the object in extsInUse if already previously created. This
```
matt-jordan, line 21: This comment is somewhat awkward. I'd rephrase it slightly:
"If this is the first time accessing this extension, then ... . Otherwise, it returns the previously created object."

lib/dal.js
```diff
@@ -2,35 +2,60 @@ var Q = require('q');
...
+ *   is done so that the busy data of the trunk will be used when 
+ *   getSharedExtension is called by another user when a call is in session.
  * @param {string} confFile - the path and filename to the configuration file
  * @param {string} name - the name of the sharedExtension to access
  * @return {Q} - Q promise object
  */
```
matt-jordan, line 28: Remove this blank line

lib/dal.js
```diff
@@ -2,35 +2,60 @@ var Q = require('q');
...
 
 var getSharedExtension = function(confFile, name) {
-  var defer = Q.defer();
   var sharedExtension = {};
-
   sharedExtension.found = false;
+
   return fetchJSON(confFile).then(function(data) {
-    data.sharedExtensions.forEach(function(extension, i) {
+    for (var index = 0; index < data.sharedExtensions.length; index++){
```
matt-jordan, line 34: Space between ')' and '{'

lib/dal.js
```diff
@@ -2,35 +2,60 @@ var Q = require('q');
...
+      var extension = data.sharedExtensions[index];
+
       if (extension[name]) {
```
matt-jordan, line 37: Invert the logic here. If extension[name] does not exist, log an error or warning and return. Then set up the sharedExtension object from the configuration.

lib/sla.js
```diff
@@ -45,80 +48,135 @@ function designateMixingBridge(client, bridgeName) {
...
+ * Updates the state of the shared extension.
  * @param {Object} client - Object that contains information from the ARI
  *   connection.
- * @param {Object} inbound - Incoming channel to be added to the bridge.
+ * @param {String} name - name of the shared extension to update.
+ * @param {String} state - state to update the shared extension to.
+ */
+function updateState(client, name, state) {
+  var deviceState = Q.denodeify(client.deviceStates.update.bind(client));
+  deviceState({deviceName: 'Stasis:' + name, deviceState: state});
```
samuelg, line 59: Why do you not return deviceState? What if an error occurred when updating the device state? As this stands there's no way for the caller to know that an error occurred.

lib/sla.js
```diff
@@ -45,80 +48,135 @@ function designateMixingBridge(client, bridgeName) {
...
  * @param {Object} dialed - Created/dialed channel to be added as well
  * @param {Object} bridge - Bridge that channels are to be added to.
+ * @param {Object} extension - the object data related to this sharedExtension
  * @return {Q} Q promise object.
  */
-function addChannelsToBridge(client, inbound, dialed, bridge) {
+function addChannelsToBridge(client, channel, dialed, bridge, extension) {
+  // The only way that the channel length will be 1 is if the outside user is
+  // still inside of the shared extension waiting for others to answer.
+  if (bridge.channels.length === 1) {
```
samuelg, line 89: Be careful when using objects that are returned as part of asynchronous calls like bridge. It reflects the state of the bridge at the time the call was made and may not reflect the current state of the bridge. It may make more sense to store this information in your application and update as your application runs.

lib/sla.js
```diff
@@ -45,80 +48,135 @@ function designateMixingBridge(client, bridgeName) {
...
+
+      } else {
+        if (object.bridge.channels.length) {
+          console.log('Hanging up station channel/s');
+
+          object.bridge.channels.forEach( function(ID) {
+
+            var hangup = Q.denodeify(client.channels.hangup.bind
+              (client.channels));
+            hangup({channelId: ID})
```
samuelg, line 154: Looks like you haven't applied the change to PR #19 that uses return Q.all(hangupArray);

lib/sla.js
```diff
@@ -45,80 +48,135 @@ function designateMixingBridge(client, bridgeName) {
...
-    }
 
-    if (numInBridge === 0) {
-      updateState(client, bridge.name, idle);
-      defer.resolve('Application completed');
-      object.bridge.removeListener('ChannelEnteredBridge', channelEntered);
-      object.bridge.removeListener('ChannelLeftBridge', channelLeft);
-    }
+      if (object.bridge.channels.length === 0) {
+        defer.resolve(true);
```
samuelg, line 164: resolve already indicates a success. Why pass a boolean to it? If you're .then function gets called after you call this function then the operation was a success. You shouldn't have to pass a boolean value to indicate that.

lib/sla.js
```diff
@@ -231,101 +332,255 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+  var defer = Q.defer();
 
-  trunkData.getAllStations()
-    .then(function (stations) {
-      if (stations.length) {
-        stations.forEach(function(station) {
-          dialed.push({endpoint: station.endpoint, channel: client.Channel()});
-        });
+  getState(client, bridge.name).then(function (currentState) {
+    if (currentState === inUse || currentState === ringing ||
```
samuelg, line 355: Move this to a utility function.

res/config.json
```diff
@@ -3,18 +3,21 @@
...
+         "SIP/kingcrimson"
+        ]
+      }
+    },
+    {
+      "420": {
+        "trunks": [
+          "42-A"
+        ],
+       "stations": [
```
samuelg, line 19: This is no aligned correctly. I would recommend prettifying the json.

lib/sla.js
```diff
@@ -45,80 +48,135 @@ function designateMixingBridge(client, bridgeName) {
...
+  });
+}
+
+/**
+ * Adds the caller channel and originated channel (if applicable) to the bridge.
+ * @param {Object} client - the client (used for hanging up at the end).
+ * @param {Object} channel - Incoming channel to be added to the bridge.
  * @param {Object} dialed - Created/dialed channel to be added as well
  * @param {Object} bridge - Bridge that channels are to be added to.
+ * @param {Object} extension - the object data related to this sharedExtension
```
matt-jordan, line 83: The description of this parameter is vague. From it, I still have no idea what an extension is, or why I need it when I add a channel to a bridge.

lib/sla.js
```diff
@@ -45,80 +48,135 @@ function designateMixingBridge(client, bridgeName) {
...
  * @return {Q} Q promise object.
  */
-function addChannelsToBridge(client, inbound, dialed, bridge) {
+function addChannelsToBridge(client, channel, dialed, bridge, extension) {
```
matt-jordan, line 86: Your extension parameter does not appear to be used in this function.

lib/sla.js
```diff
@@ -45,80 +48,135 @@ function designateMixingBridge(client, bridgeName) {
...
+  // Resets listeners, so that redundant events aren't accepted.
+  bridge.removeAllListeners('ChannelEnteredBridge');
+  bridge.removeAllListeners('ChannelLeftBridge');
   var defer = Q.defer();
-  var numInBridge = 0;
 
   console.log('Adding channels to bridge %s', bridge.id);
 
   var addChannel = Q.denodeify(bridge.addChannel.bind(bridge));
-  addChannel({channel: [inbound.id, dialed.id]})
```
matt-jordan, line 63: Rather than treating your inbound and dialed channel separately, have the function take in a list of channels to add. Then add all of them iteratively (which the underlying API does support).

Callers of this function should build a list and call it with that list - they can make the determination of whether or not a dialed channel can be added or not.

tests/sla_tests.js
```diff
@@ -619,5 +709,387 @@ describe('SLA Bridge and Channels Tester', function() {
...
+
+    var sla = require('../lib/sla.js')(client, config, channel, '42')
+      .catch(errHandler)
+      .done();
+
+    extensionUnhold();
+
+    function extensionUnhold() {
+      setTimeout(function() {
+        if (channels.length === 2 && bridgeChannels.length === 2 &&
```
samuelg, line 1068: The next ~25 lines could use vertical spacing.

lib/sla.js
```diff
@@ -45,80 +48,135 @@ function designateMixingBridge(client, bridgeName) {
...
       defer.reject(err);
     });
 
+  if (dialed) {
+    addChannel({channel: dialed.id})
+      .catch(function (err) {
+        defer.reject(err);
+      });
+  }
+  bridge.on('ChannelEnteredBridge', channelEntered);
```
matt-jordan, line 113: Should you be subscribing for this event on every single channel that enters? Or should it simply be done when the bridge is created?

lib/sla.js
```diff
@@ -45,80 +48,135 @@ function designateMixingBridge(client, bridgeName) {
...
+  bridge.on('ChannelLeftBridge', channelLeft);
+  return defer.promise;
+
+  /**
+   * Function that displays a helpful message when a channel enters a bridge
+   * @param {Object} event - the event related to this bridge entering
+   * @param {Object} object - contains the bridge and channel in question
+   */
   function channelEntered(event, object) {
+    object.channel.on('ChannelHold', channelHold);
```
matt-jordan, line 123: This dual-purposes this function. If its purpose is to display a helpful message when a channel enters a bridge, it should have nothing to do with handling or responding to a channel hold event.

lib/sla.js
```diff
@@ -45,80 +48,135 @@ function designateMixingBridge(client, bridgeName) {
...
-          });
-      }
-      else if (object.channel.id === dialed.id) {
-        console.log('Hanging up inbound channel %s', inbound.id);
-        var hangupInbound = Q.denodeify(inbound.hangup.bind(inbound));
-        hangupInbound()
-          .catch(function (err) {
-            err.name = 'HangupFailure';
-            defer.reject(err);
+    object.channel.removeAllListeners('ChannelHold');
```
matt-jordan, line 137: Again, removing or modifying listeners for hold when a channel leaves a bridge doesn't feel correct. You should do that when the channel leaves your application; the bridge is ancillary to that.

lib/sla.js
matt-jordan, line 123: You also should not be handling a ChannelHold event from every channel that enters the bridge. You should only be intercepting hold from stations.

lib/sla.js
```diff
@@ -45,80 +48,135 @@ function designateMixingBridge(client, bridgeName) {
...
+
+    isStation(extension, object.channel).then(function (isStation) {
+      if (isStation) {
+        if (object.bridge.channels.length === 1) {
```
matt-jordan, line 141: Two issues with this:

1) You aren't checking to see if the last channel in the bridge is a station. Right now, this will start playing MoH to another station, which isn't correct.
2) You aren't checking that a station left because it pressed hold. It looks like this routine is called any time a channel leaves a bridge. They can leave a bridge because they hung up, in which case, you shouldn't play MoH to the trunks - they aren't on hold, they were hung up on. In that case, they should probably be ejected from the shared extension.

lib/sla.js
```diff
@@ -45,80 +48,135 @@ function designateMixingBridge(client, bridgeName) {
...
-    }
+      if (object.bridge.channels.length === 0) {
+        defer.resolve(true);
+        updateState(client, bridge.name, idle);
+        bridge.removeAllListeners('ChannelEnteredBridge');
+        bridge.removeAllListeners('ChannelLeftBridge');
+      }
+    });
+  }
+  function channelHold(event, channel) {
```
matt-jordan, line 171: Space between functions please. Also document your functions.

lib/sla.js
```diff
@@ -45,80 +48,135 @@ function designateMixingBridge(client, bridgeName) {
...
+    isStation(extension, channel).then(function (isStation) {
+      if (isStation) {
```
matt-jordan, line 173: You shouldn't have to check if this is a station or not, since you control whose holds you intercept based on who you put the function on.

app.js
```diff
@@ -34,6 +35,7 @@ function clientLoaded (client) {
...
       sla(client, confFile, channel, extension)
```
matt-jordan, line 39: Somewhere in your SLA function I would expect to see the HOLD_INTERCEPT function be applied to the channel. I haven't found the application of that dialplan function on a channel in this review; as it is, I'm not sure how you're intercepting hold events without it.

lib/sla.js
```diff
@@ -128,13 +186,58 @@ function updateState(client, name, state) {
...
  *   addChannelsToBridge().
- * @ param {Object} trunkData - the data of the trunk to be used for station
- *   access.
+ * @param {Object} extension - the data structure that represents the extension,
+ *   mainly used for telling whether or not an extension is busy or not
  * @return {Q} Q promise object.
  */
-function originateChannel(client, inbound, bridge, trunkData) {
-  var defer = Q.defer();
+function originateInbound(client, inbound, bridge, extension) {
```
matt-jordan, line 193: Why is this named originateInbound? Generally, if you are originating a channel, you are creating an outbound channel from Asterisk to some device. The name here feels confusing.

lib/sla.js
```diff
@@ -128,13 +186,58 @@ function updateState(client, name, state) {
  * @param {Object} inbound- the inbound channel
  * @param {Object} bridge - Object that defines the bridge to be passed to
  *   addChannelsToBridge().
- * @ param {Object} trunkData - the data of the trunk to be used for station
- *   access.
+ * @param {Object} extension - the data structure that represents the extension,
```
matt-jordan, line 189: The comment here is misleading. extension is used to obtain the stations to place an outbound call to, not to determine their state.

lib/sla.js
```diff
@@ -128,13 +186,58 @@ function updateState(client, name, state) {
...
+          if (stations.length) {
+            stations.forEach(function(station) {
+              dialed.push({endpoint: station, channel: client.Channel()});
+            });
+
+            dialed.forEach(function (object) {
+              var originate = Q.denodeify(object.channel.originate.bind(
+                  object.channel));
+              originate({endpoint: object.endpoint, app: 'sla',
+                  appArgs: 'dialed', timeout: 10})
```
matt-jordan, line 214: Your timeout parameter should be configurable. I'd go ahead and put that as an item on the project plan.

lib/sla.js
```diff
@@ -128,13 +186,58 @@ function updateState(client, name, state) {
...
-function originateChannel(client, inbound, bridge, trunkData) {
-  var defer = Q.defer();
+function originateInbound(client, inbound, bridge, extension) {
+
   var dialed = [];
+  var defer = Q.defer();
+
+  getState(client, bridge.name).then(function(currentState) {
+
+    if (currentState !== inUse && currentState !== busy &&
```
matt-jordan, line 200: Invert the logic: if the bridge is not in the correct state, return early, and reduce indentation.

lib/sla.js
```diff
@@ -128,13 +186,58 @@ function updateState(client, name, state) {
...
+                });
+
+              updateState(client, bridge.name, ringing);
+
+              object.channel.once('StasisStart', dialedEnteredStasis);
+              object.channel.once('ChannelDestroyed', dialedChannelHungup);
+            });
+           inbound.once('ChannelHangupRequest', inboundHungup);
+          } else {
+            defer.reject(new customError.CustomError('NoStations',
```
matt-jordan, line 226: I would expect that a shared extension must have at least one station, and this this should be caught earlier in the process - probably in the DAL.

lib/sla.js
```diff
@@ -128,13 +186,58 @@ function updateState(client, name, state) {
...
+
+  getState(client, bridge.name).then(function(currentState) {
+
+    if (currentState !== inUse && currentState !== busy &&
+        currentState !== ringing) {
+      extension.getAllStations()
+        .then(function (stations) {
+
+          if (stations.length) {
+            stations.forEach(function(station) {
```
matt-jordan, line 206: I could be wrong, but the extra creation of the dialed list feels extraneous. I would expect that you could do this as:

```
stations.forEach(function (station) {
    var channel = client.Channel()
    var originate = Q.denodeify(channel.originate.bind(channel))
    originate({endpoint: station, app: 'sla', appArgs: 'dialed', timeout: 10})
```

Something like that.

lib/sla.js
```diff
@@ -128,13 +186,58 @@ function updateState(client, name, state) {
...
+                .catch(function (err) {
+                  defer.reject(err);
+                });
+
+              updateState(client, bridge.name, ringing);
+
+              object.channel.once('StasisStart', dialedEnteredStasis);
+              object.channel.once('ChannelDestroyed', dialedChannelHungup);
+            });
+           inbound.once('ChannelHangupRequest', inboundHungup);
```
matt-jordan, line 224: How does this hang up your dial attempts to the stations?

lib/sla.js
```diff
@@ -128,13 +186,58 @@ function updateState(client, name, state) {
...
+          } else {
+            defer.reject(new customError.CustomError('NoStations',
+                  'No stations in this shared extension.'));
+          }
+        });
+    } else {
+      // This will either redirect this channel towards a Hangup or a "backup
+      // extension"
```
matt-jordan, line 232: Add an item to your wiki project plan to set a channel variable if an inbound trunk is bounced out of a shared extension due to it being already in use by another trunk.

lib/sla.js
matt-jordan, line 200: Move the state logic checks to a utility function.

lib/dal.js
wolfeman67, line 37: extension[name] is used to check if there is an extension with the name "name". You search through the data's sharedExtensions and check if they have a property "name". For example, a shared extension of name "42" would check if an extension in the array has a property of "42", and if so return that object. I do not think it would make sense to log an error for every extension that doesn't have that particular name, but I could be incorrect.

lib/sla.js
wolfeman67, line 89: I may be incorrect, but I think the only way that the bridge.channels.length could only be equal to one is if the outside channel was left hanging on hold. When an initial inbound or outbound call is made, two channels are always involved (insider and outsider). From there, only one station at a time can be added. So if there is only one channel left hanging, it has to be an outside channel. No outside channel means no call going on.

res/config.json
wolfeman67, line 19: Perhaps I'm using the wrong prettifier- http://jsonformatter.curiousconcept.com/ - but it looks the same as what I had originally.

lib/sla.js
wolfeman67, line 86: It's used in the event handler functions inside addChannelsToBridge (those appear after the main function body code).

lib/sla.js
wolfeman67, line 63: This may be the less efficient way of adding stations into the call; but if there is a call occurring, a station joins the extension by itself (it's set to the channel variable), and dialed is set to null. With how Javascript works, I cannot add an array of channels (with their id's) if a particular channel is undefined (has no id property). We originally had it as an array, but I thought that this was the more acceptable approach for when I only add one channel.

lib/sla.js
wolfeman67, line 113: Would it be a good idea to delete the bridge when the extension is "cleared"? This would fit with what I think you're trying to get at. Without that, there may be a chance that eventListeners would not be added at all, or that they would be added an extra time (from when I reuse a mixing bridge).

lib/sla.js
wolfeman67, line 141: 1) I could be incorrect in my reasoning, but I think with how the code works, the only way that there could be one channel left over is if the channel is a non-station. If a non-station hangs up and there is two stations remaining, both stations will be hungup at the same time (with Q.all) and this conditional will either be ignored, or the application will bypass this so fast and go to the NOT_INUSE conditional and the user won't notice anything.
2) This has been fixed, to some extent. It doesn't determine which user pressed their hold key, it just checks whether or not hold was pressed. If station A presses hold and station B hangs up, either-or could take care of the user on hold. If this seems to be a bad idea to you, I could take a different approach with this.

app.js
wolfeman67, line 39: To my knowledge, channel.on('ChannelHold') is all that you need in order to get those events. I am not entirely sure what you mean by applying the HOLD_INTERCEPT function either.

lib/sla.js
wolfeman67, line 214: Should I have a default value for this, just in case a user doesn't want to input it?

lib/sla.js
wolfeman67, line 206: The dialed array is mainly used for the future hangups of channels when an inbound user hangs up, or if a station picks up and the rest have to be hung up (only 1 station gets the call initially). I don't remember exactly what it was for, but Sam suggested making the dialed object contain both the station endpoint and the channel related to that station. I think dialed is necessary, but I could be incorrect.

lib/sla.js
wolfeman67: I think I did so under an item of Phase 2 (Busy Extensions/Multiple Stations): 
"If another outside user tries to come into a sharedExtension that is currently in use, they will continue going through the rest of the dialplan, so that it can either A. go to another shared extension, or B. get hung up gracefully."

lib/sla.js
wolfeman67, line 224: When an inbound channel hangs up, it goes to the inboundHungup function, and loops through all of the channels in dialed, hanging each up in the process. inboundHungup is an eventListener function that appears after the main body code. Sam said that the acceptable way of writing Javascript code was writing eventListener functions (or other similar functions) after the main code.

samuelg, line 224: The issue was that code that setup event listeners, code that ran before an event listener fired, and the actual event listeners were interspersed at the same level which made it difficult to read. It's better to group related code together. I prefer to put event listener implementation at the bottom when using embedded functions but it's a stylistic preference; not necessarily acceptable vs unacceptable.

lib/sla.js
samuelg, line 206: I believe dialed is used to keep track of stations/channels throughout the lifetime of the application for a given extension.

Comments would definitely help here.

lib/sla.js
matt-jordan, line 86: Which line number? I'm still missing it.

Even if it _is_ used in this function, this feels like a function that is starting to get overloaded with too many purposes. Adding channels to a bridge should not be responsible for more than simply adding the channels to the bridge. I'm not sure why an extension should be part of the purpose of this function.

lib/sla.js
matt-jordan, line 89: I don't agree with that assumption. You should not expect that you only have a station left in the shared extension.

Consider the following scenarios:
- Station joins the SE (channels == 1)
- Station calls a trunk (channels == 2)
- Station and trunk talk. Station hits hold, causing them to leave the bridge (channels == 1)
- trunk stays in SE, and hears hold music.

Another scenario:
- Station 1 joins the SE (channels == 1)
- Station 2 joins the SE (channels == 2)
- Trunk joins the SE (channels == 3)
- Trunk hangs up (channels == 2)
- Station 1 hits hold, disconnecting
- Stations 2 does _not_ hear MoH

Both of these scenarios are mentioned in the issue requirements.

RIght now, you have nothing that I can find that keeps track of the types of channels in the SE, and makes use of that knowledge when the channel count changes. I would expect to see some data structure holding the channels and their types in the bridge, and when the state changes, alters the state of the bridge accordingly based on that information.

lib/sla.js
matt-jordan, line 63: Why would you not add a list of length 1 (one channel) in that scenario?

lib/sla.js
matt-jordan, line 113: What I'm saying isn't tied to the lifetime of a bridge. You can subscribe for events when the bridge is created without also destroying the bridge later.

I would expect that if I wanted to handle an event every time it happened, I would subscribe once for that event. Subscribing multiple times creates additional overhead that is unnecessary.

lib/sla.js
matt-jordan, line 141: I addressed point #1 up above. Your requirements do not state that you can't have two stations talking to each other. If that's how your implementation works, then your implementation is wrong.

If a station hangs up, and only a trunk is left in the SE, then they shouldn't hear hold music. That means you need to have logic that keeps track of why a channel left the bridge (either hold or hangup), and makes the appropriate decisions.

lib/sla.js
matt-jordan, line 206: I would also rename that array, if that is its intended purpose.

lib/sla.js
matt-jordan, line 214: Sure.

lib/sla.js
wolfeman67, line 86: It's used by the isStation function which checks whether or not inside of the sharedExtension data, a given channel is a station or not. This function is used on lines 139 and 172. I suppose what I could do is extract the array of stations before this function is used, if that sounds reasonable.

README.md
superscrat, line 35: When you are describing abstract ideas, it is better to break the ideas into small, concise thoughts. Using too many pronouns can convolute your intention and become confusing to the reader. Consider the following (You would want to keep the original paragraph form; I put each sentence on its own line only for demonstration):

You must also have a dialplan extension in extensions.conf that leads to the application.
[The dialplan extension] must have same name as what application is being started in the code.
[The dialplan extension] must have an argument to represent the SLA bridge to reach.
An example with a device state hint is below:

These thoughts are easier to follow than long, winding sentences. That sort of style seems to be good for writing prose but not effective when writing technical documentation.

app.js
superscrat, line 12: I agree with Matt's comment. When you are building software, remember to keep the changes small and targeted. It is a lot easier to find the source of a regression if you only change one variable at a time.

lib/dal.js
superscrat, line 37: I think you should log something, even if the message is classified as informational. The extra messages will help the next person who has to debug this.

lib/sla.js
superscrat, line 113: What Matt is saying, is by subscribing to the event here, on line 113, rather than in some other place where the code will only be executed once, will result in the list of subscribers growing inordinately, which may have unexpected side-effects the longer the application is alive.

Consider the relatively straightforward mechanics of firing an event: the event has some collection of subscribers (callbacks) that are invoked when the event is raised. 

The event does not care if {n} callbacks are identical; it is going to treat each request for subscription as an individual and unique request. In essence, each time line 113 is executed, the event appends the callback to its list of subscribers. So, if this code is executed 18 times, whenever the event, 'ChannelEnteredBridge' is raised, your callback will be also executed 18 times. 

I have a strong hunch this is not what you are wanting.

res/config.json
superscrat, line 19: I think it's your tab size is 3 rather than 4. The asterisk guidelines state that the tab size should be 4: https://wiki.asterisk.org/wiki/display/AST/Coding+Guidelines#CodingGuidelines-Codeformatting

tests/sla_tests.js
```diff
@@ -40,8 +40,14 @@ var config = 'tests/testConfigs/singleEndpoint.json';
 var mockDeviceStates = [];
 // Device state object
 var ds = {};
+// Whether or not moh has started on the bridge
+var mohStarted = false;
```
superscrat, line 44: I am confused why you need 2 variables to denote the same thing.

If mohStarted = false, does that not imply that mohStopped = true?

I do not like the notion of maintaining lots of state variables - it always becomes cumbersome and difficult to unwind when it is time to revisit the code.

tests/sla_tests.js
```diff
@@ -186,25 +243,52 @@ var getMockChannel = function() {
...
           });
+
           self.emit('ChannelHangupRequest', {channel: self}, self);
           self.emit('ChannelDestroyed', {channel: self}, self);
           cb(null);
         }
       }, (asyncDelay/2));
     };
+
+    this.play = function(media, playbackObj) {
```
superscrat, line 278: What is the purpose of this.play? All it does is return null. It seems like an overkill way to assign null to some object.

tests/sla_tests.js
```diff
@@ -282,8 +374,8 @@ describe('SLA Bridge and Channels Tester', function() {
     function validChecking() {
       setTimeout(function() {
         if (dialed[0] && dialed[0].wasAnswered && bridgeChannels.length !== 0 &&
-            isMixing && channel.wasAnswered) {
-              done();
+          isMixing && channel.wasAnswered) {
```
superscrat, line 377: I am not sure I am a fan of the 'was' verb in this variable. I think channel.answered would be sufficient.

Also, I keep seeing loads of state variables. I think that you should spend some time trying to consolidate these (if possible). 

The way to write beautiful code is to in fact, write less code. 

The less code you have, the fewer data paths you have to unwind in the event of fixing bugs or extending functionality. It is also much easier for a human to easily glean intent without having to execute the code through a debugger.

tests/sla_tests.js
```diff
@@ -619,5 +709,387 @@ describe('SLA Bridge and Channels Tester', function() {
...
+              .catch(errHandler)
+              .done();
+            }, asyncDelay);
+          }
+          extensionUnhold();
+        }
+      }, asyncDelay);
+    }
   });
 });
```
superscrat, line 1095: For a test, 1095 lines of code is encroaching upon code smell. I think you should investigate ways to break this file up into smaller, more modular components. 

Having too much in one file will cause bloat and make maintenance very challenging.

As a starting place, look for areas where you have almost (if not exactly the same) code. Extract those lines into a function. Remember, one of the first principles is "do not repeat yourself".

res/config.json
samuelg, line 19: 2 is more common in JavaScript code bases. They are more than welcomed to use 4 of course, but figured I would mention it since my coding style probably has influenced them but in this case the indenting is incorrect regardless.

wolfeman67:
The stuff in HoldBasic has been reworked inside of hangupHoldLogic. A lot of the problems inside of this PR were also related to outboundUno so I just combined what I had in this PR, and outboundUno's PR, and removed the hold logic of that combined PR, namely, extensionObjects. extensionObjects also has a data object which represents the current trunk and stations of all of the sharedExtensions, so that the application has direct control of what is going on.

PR #26, opened 2015-04-23
Outbound single (Syntax only)

wolfeman67:
This is the simple whitespace and grammatical fixes of outbound single.  All addressed findings of the original outboundSingle can be found at PR #25.


lib/sla.js
```diff
@@ -45,80 +49,113 @@ function designateMixingBridge(client, bridgeName) {
...
-    console.log('Channel %s has left the bridge', object.channel.id);
-
-    if (numInBridge === 1) {
-      if (object.channel.id === inbound.id) {
-        console.log('Hanging up dialed channel %s', dialed.id);
-        var hangupDialed = Q.denodeify(dialed.hangup.bind(dialed));
-        hangupDialed()
+    if (isStation(extension, object.channel)) {
+      if (object.bridge.channels.length === 1) {
+        var hangupOutsider = Q.denodeify(client.channels.hangup.bind
```
samuelg: This was not addressed. As Ashley pointed out, this variable name is confusing and does not really indicate what this is hanging up by the addition of the name 'Outsider'. I would recommend calling it hangup to match the name of the function being denodeified.

Also, since you are modifying this, binding to client.channels will not work. You should bind to client. Since you have a channel instance however (object.channel), you can just bind to that and then you won't have to pass a channelId parameter to the call.

My guess is that this did not work so you added a channelId parameter to get around the issue.

lib/sla.js
```diff
@@ -231,101 +305,251 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+    }
+  });
   return defer.promise;
+  /**
+   * The function that gets called once the dialed channel enters Stasis.
+   *   Makes the extension busy and clears eventListeners
+   * @param {Object} event - the information related to the StasisStart event
+   * @param {Object} line - the line that has entered Stasis
+   */
+  function dialedEnteredStasis(event, line) {
```
samuelg: This was not addressed. You still have 2 functions with the same name which is confusing. Rename them to better reflect the difference between the 2.

lib/sla.js
```diff
@@ -231,101 +305,251 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
+    var answer = Q.denodeify(line.answer.bind(line));
+    answer();
+    defer.resolve(line);
+  }
+
+  /**
+   * Function that gets called when a dialed channel hangs up
+   * @param {Object} event - the event object related to this hang up
+   */
+  function dialedChannelHungup(event) {
```
samuelg: This was not addressed. You still have 2 functions with the same name which is confusing. Rename them to better reflect the difference between the 2.

samuelg: Ashley also mentioned renaming the event handlers to make it more clear that they are event handlers. This was also not addressed. Although here I would suggest not using on_event_name and instead using onEventName to match the camel case used throughout the project.

lib/sla.js
```diff
@@ -231,101 +305,251 @@ function originateChannel(client, inbound, bridge, trunkData) {
...
-      return Q.reject(new customError.CustomError('InvalidTrunkSpecification',
-          invalid));
-    });
-  })
-  .catch(function(invalid) {
-    if (invalid.name === 'InvalidSpecification') {
-      return Q.reject(invalid);
+function isStation (extension, channel) {
+  var stations = extension.getAllStations();
+  var isItStation = false;
```
samuelg: isItStation is rather awkward. Since this is the result of the function itself you can simply call it result.

samuelg:
I am done going through the whitespace/syntax findings from PR #19.

samuelg:
Looks good. Go ahead and squash the commits and merge to master. You can then rebase the branch PR 25 is based off of and we'll go from there.

wolfeman67:
Squashed into #27.

